PR #6, opened 2017-08-22
Allow insecure curl requests

sadatay:
Since okapi.frontside.io is now secured with ssl, we have to explicitly tell curl to skip certificate verification or the module registration and discovery will fail.


cowboyd:
Why can't curl verify the certificate?

PR #11, opened 2017-09-21
Support JSON-API configuration endpoint

sadatay:
To gain access to a Knowledge Base ie. EBSCO you need to have valid credentials which are stored as `configuration` records. As there can be an infinite number of end-users all having unique credentials to the Knowledge Base we need to allow users the ability to create credentials unique to them. Here we implement a new `eholdings/configuration` endpoint to enable the capability of manipulating those credentials. This endpoint currently has the ability to `POST`, `GET`, and `DELETE` credentials. Here we treat the `configuration` records as immutable objects and manipulating Knowledge Base credentials will in turn return a brand new `configuration` record with the updated credentials.

All requests that are not to the `eholdings/configuration` endpoint function in the way they have function thusfar - passing through directly to the RM-API.


cowboyd:
Have we evaluated other serialization layers like `jsonapi-resources` or `jsonapi-rb` ?

app/resources/serializable_configuration_record.rb
```diff
@@ -0,0 +1,19 @@
+class SerializableConfigurationRecord < JSONAPI::Serializable::Resource
+  type 'configuration'
+
+  attribute :'api-key' do
+    @object.api_key
+  end
+
+  attribute :'customer-id' do
+    @object.customer_id
+  end
```
sadatay: wish there was a better way to do this but i haven't found one yet.  if there isn't, that's one strike against `jsonapi-rails` in my book, considering the json-api definition [recommends](http://jsonapi.org/recommendations/) dashes

cowboyd: I agree, that stinks.

sadatay:
imo we should dump `jsonapi-rb`.  besides the whole snake/kebab thing, it looks like the `meta` key is sunk into the `data` key, which doesn't match the [spec](http://jsonapi.org/examples/#pagination):

screen shot 2017-09-22 at 11 38 52 am

cowboyd:
I'm ok with this.

app/models/application_model.rb
```diff
@@ -0,0 +1,18 @@
+class ApplicationModel
```
cowboyd: I don't believe we need this class anymore.

app/models/configuration.rb
```diff
@@ -0,0 +1,49 @@
+require "net/http"
+require "uri"
+require "json"
+
+class Configuration
```
cowboyd: We need some documentation around this class.

app/models/configuration_record.rb
```diff
@@ -0,0 +1,6 @@
+class ConfigurationRecord < ApplicationModel
```
cowboyd: I don't believe this is necessary anymore.

app/resources/configuration_resource.rb
```diff
@@ -0,0 +1,18 @@
+require 'okapi'
+
+class ConfigurationResource < JSONAPI::Resource
+  attributes :customer_id, :api_key
+
+  def self.find_by_key(key, options = {})
```
cowboyd, line 6: We need some documentation around this method (i.e. it's necessary to implement singular, not active-record based resources), preferably a link to the issue on github that outlines this solution.

app/resources/configuration_resource.rb
```diff
@@ -0,0 +1,18 @@
...
+    context = options[:context]
+    okapi = context[:okapi]
+    model = Configuration.new(okapi)
+    model.load!
+
+    resource_for_model(model).new(model, context)
+  end
+
+  def model_error_messages
```
cowboyd, line 15: We should document what this is, and how it fits into the jsonapi resource request cycle.

app/resources/serializable_base.rb
```diff
@@ -0,0 +1,11 @@
+class SerializableBase < JSONAPI::Serializable::Resource
```
cowboyd: I don't think we need this anymore, this is a jsonapi-rb thing. Now that we're on jsonapi-resources

app/resources/serializable_configuration_record.rb
```diff
@@ -0,0 +1,13 @@
+class SerializableConfigurationRecord < SerializableBase
```
cowboyd: Also not needed. This surfaces something though. We should have the validity check being in the meta data. I don't think we have a test for that.

cowboyd (changes requested):

app/resources/serializable_base.rb
elrickvm: true

app/models/configuration_record.rb
elrickvm: true

app/models/application_model.rb
elrickvm: true

app/controllers/api/v1/api_controller.rb
```diff
@@ -0,0 +1,34 @@
+require 'net/http'
+
+module Api
```
elrickvm: Should this be remove from being nested under v1? Can move all of this to app/controllers/application_controller.rb?

app/controllers/application_controller.rb
```diff
@@ -1,2 +1,29 @@
+require 'net/http'
 class ApplicationController < ActionController::API
```
cowboyd: Now that we have more than one controller, I'm not sure how comfortable I am with all of this logic going into application controller which is supposed to be a shared super class for all controllers.  The expectation in rails is that unless otherwise specified, you would inherit from `ApplicationController`.

What about `RMAPIProxyController` since that's effectively what happens.

app/controllers/application_controller.rb
```diff
@@ -1,2 +1,29 @@
...
+  def index
+    # This is currently just a passthrough to the RM-API
+
+    # Form the URL
+    urlstr = "#{ENV['EBSCO_RESOURCE_MANAGEMENT_API_BASE_URL']}/rm/rmaccounts/#{ENV['EBSCO_RESOURCE_MANAGEMENT_API_CUSTOMER_ID']}#{request.fullpath}"
```
cowboyd: We actually need to use our configuration here. We'll need to have the same logic to verify the okapi headers.

cowboyd (changes requested):
We're writing the configuration, but we're still using an environment variables to connect to the RMAPI.

Let's refactor the okapi verification into the application controller, and then take the proxy logic and put it into its own controller.

app/models/configuration.rb
```diff
@@ -0,0 +1,49 @@
...
+    end
+   response
+  end
+
+  def id
+    'rmapi'
+  end
+
+  def valid?(arg)
+    urlstr = "https://sandbox.ebsco.io/rm/rmaccounts/#{@customer_id}/vendors?search=zz12&offset=1&orderby=vendorname&count=1"
```
sadatay: i'm guessing we're going to continue using an ENV var for the RMAPI address for the time being.  we should probably use that here instead of hardcoding

config/routes.rb
```diff
@@ -1,3 +1,6 @@
 Rails.application.routes.draw do
-  match '*path' => 'api#index', via: [:get, :post, :put, :patch, :delete]
+  scope '/eholdings' do
+    jsonapi_resource :configuration, only: [:show, :update]
```
sadatay, line 3: a routing helper!  la-ti-da `jsonapi::resources`

app/models/configuration.rb
```diff
@@ -0,0 +1,49 @@
+require "net/http"
+require "uri"
+require "json"
```
cowboyd: I think all we need here are "open-uri"

cowboyd (approved):

PR #13, opened 2017-10-19
Implement status endpoint

wwilsman:
- Uses the configuration model for checking validity.
- Extends the JSON API controllers from a base controller.
- Updates the modules descriptor and increments the version for `mod-kb-ebsco`


wwilsman:
Depends on #11

PR #15, opened 2017-10-19
Explicitly return true from validation

cowboyd:
Otherwise, the validation flag is the actual stream of the http response to the EBSCO RMApi


app/models/configuration.rb
```diff
@@ -27,7 +27,7 @@ def valid?(arg = {})
     urlstr = "#{ENV['EBSCO_RESOURCE_MANAGEMENT_API_BASE_URL']}/rm/rmaccounts/#{@customer_id}/vendors?search=zz12&offset=1&orderby=vendorname&count=1"
 
     open(urlstr,{'X-Api-Key' => @api_key})
-
+    return true
```
sadatay, line 30: not chill

sadatay (approved):

PR #16, opened 2017-10-21
Use jsonapi-rb For Resource Serialization

sadatay:
`jsonapi-resources` let us down a bit, so this reverts to the much leaner `jsonapi-rb` via `jsonapi-rails`.  There's some general tidying as well.  For vital ENV vars I've switched to using the `fetch` method, which will explicitly fail instead of resolving to `nil`.  I tried to tidy up some styling wherever I saw the opportunity.  

~~Tests are passing but I'll be manually beating the crap out of this thing via telepresence so that we are _assured_ it will behave itself in the cluster.~~
Testing in the cluster revealed some additional bugs.  I've addressed them, and the changes I pushed tonight stood up against everything I threw at it.  Also went ahead and bumped the ruby version to 2.4.2.


app/controllers/proxy_controller.rb
```diff
@@ -1,17 +1,15 @@
 class ProxyController < ApplicationController
-  def configuration
-    @config ||= ::Configuration.new(okapi).tap do |config|
-      config.load!
-    end
-  end
+  before_action :slice_eholdings_namespace
```
cowboyd: Let's compute the desired path with a simple function instead of an action that, as one of its side effects, mutates `request.fullpath`

app/controllers/proxy_controller.rb
```diff
@@ -31,4 +29,18 @@ def index
 
     render :json => response.body, :status => response.code
   end
+
+  private
+
+  def slice_eholdings_namespace
```
cowboyd: How about just

```
def rmapi_path
  request.fullpath.slice('/eholdings')
end
```

And then just use that in the template string directly.

app/controllers/proxy_controller.rb
```diff
@@ -1,17 +1,15 @@
...
-
-    # Form the URL
-    urlstr = "#{ENV['EBSCO_RESOURCE_MANAGEMENT_API_BASE_URL']}/rm/rmaccounts/#{configuration.customer_id}#{request.fullpath}"
-    urlstr.slice! '/eholdings'
-    uri = URI(urlstr)
+    # For the time being, this is a direct passthrough to
+    # the EBSCO RM-API.  As more of Codex is incorporated, this will
+    # be replaced with more robust serialization.
+    uri = URI(
+      "%s/rm/rmaccounts/%s%s" % [
```
cowboyd: I'm not opposed to using this method of substitution instead of a template string, but if we go this route, let's use named keys instead of positional:

```ruby
"%{base}/rm/rmaccounts/%{customer}%{path}" % base: rmapi_base_url, customer: @customer_id, path: request.fullpath
```

app/models/configuration.rb
```diff
@@ -1,33 +1,45 @@
...
+    end
   end
 
-  def valid?(arg = {})
-    urlstr = "#{ENV['EBSCO_RESOURCE_MANAGEMENT_API_BASE_URL']}/rm/rmaccounts/#{@customer_id}/vendors?search=zz12&offset=1&orderby=vendorname&count=1"
 
-    open(urlstr,{'X-Api-Key' => @api_key})
-    return true
+  def valid?
+    uri = "%s/rm/rmaccounts/%s/vendors?search=zz12&offset=1&orderby=vendorname&count=1" %
```
cowboyd: Again, I think that named substitutions are easier to read and understand

cowboyd (changes requested):
Looking good. A few nitpicks, but a great improvement in readability and also more on the beaten path of a rails application.

app/controllers/proxy_controller.rb
sadatay: my original intent was to set this to an instance variable, but i guess i never got around to correcting that.  your proposed approach is better either way, so i'll go that route.

app/controllers/proxy_controller.rb
sadatay: i didn't know you could use named keys in format strings!  that's excellent

app/controllers/proxy_controller.rb
sadatay: this is what i thought you meant in your above comment.  it breaks us free of the unnecessary `before_action`.  it needs a slight tweak from how you've written it (`.slice` returns the sliced segment not the remaining segment) but this is the right idea for sure.

cowboyd (approved):

sadatay:
~~⚠️  manual testing has uncovered some bugs with this.  do not merge for now~~

PR #17, opened 2017-10-31
Implement JSON API endpoint for vendors

wwilsman:
## Purpose
Implements a `/eholdings/jsonapi/vendors` endpoint for consuming vendor resources as JSON API documents.

## Approach
- Uses `jsonapi:serializable` to serialize a vendor resource as JSON API
- When searching for vendors, the `search` parameter is now `q`
- When searching for vendors, `totalResults` is now available in the response meta
- Transforms important vendor attributes to be JSON API compliant

  **before:**
  ```
  {
    vendorId: 19,
    vendorName: "Ebsco",
    packagesTotal: 602,
    packagesSelected: 48,
    vendorToken: null,
    isCustomer: false
  }
  ```
  **after:**
  ```
  {
    data: {
      id: "19",
      type: "vendors",
      attributes: {
        name: "Ebsco",
        packagesTotal: 602,
        packagesSelected: 48
      }
    }
  }
  ```
- Any errors returned from the RM API are simply transformed

  **before:**
  ```
  {
    Errors: [
      {
        Code: 1001,
        Message: "RM API error message"
        SubCode: 0
      }
    ]
  }
  ```
  **after:**
  ```
  {
    errors: [
      {
        title: "RM API error message"
      }
    ]
  }
  ```

### DIFF
--- a/app/controllers/vendors_controller.rb
+++ b/app/controllers/vendors_controller.rb
@@ -1,0 +1,85 @@
+class VendorsController < ApplicationController
+  def index
+    uri = URI(rmapi_path)
+
+    # Transform query params to what the EBSCO RM-API expects
+    uri.query = URI.encode_www_form(
+      search: params[:q],
+      orderby: params[:orderby] || params[:q] ? 'relevance' : 'vendorname',
+      count: params[:count] || 25,
+      offset: params[:offset] || 1
+    )
+
+    # Make the RM-API request and parse the response body
+    response = get_resource(uri)
+    data = Map JSON.parse(response.body)
+
+    # Return the list of vendors
+    if response.message === 'OK'
+      vendors = data.vendors.map do |vendor|
+        Vendor.new(vendor)
+      end
+
+      render jsonapi: vendors,
+             meta: { totalResults: data.totalResults },
+             status: :ok
+    else
+      render_rmapi_errors(data, response.code)
+    end
+  end
+
+  def show
+    uri = URI(
+      "#{rmapi_path}/%{vendor_id}" % {
+        vendor_id: params[:id]
+      }
+    )
+
+    # Make the RM-API request and parse the response body
+    response = get_resource(uri)
+    data = Map JSON.parse(response.body)
+
+    # Return the requested vendor
+    if response.message === 'OK'
+      vendor = Vendor.new(data)
+      render jsonapi: vendor, status: :ok
+    else
+      render_rmapi_errors(data, response.code)
+    end
+  end
+
+  def get_resource(uri)
+    # Create the HTTP object
+    http = Net::HTTP.new(uri.host, uri.port)
+    http.use_ssl = true
+
+    # Create and send the request
+    http.get(
+      uri.request_uri,
+      {
+        "X-Api-Key" => configuration.api_key,
+        "Content-Type" => 'application/json',
+        "Accept" => 'application/json'
+      }
+    )
+  end
+
+  def render_rmapi_errors(data, code)
+    # Map RMAPI errors to JSON-API error objects
+    jsonapi_errors = data.Errors.map { |err| { title: err.Message } }
+    render jsonapi_errors: jsonapi_errors, status: code
+  end
+
+  def rmapi_path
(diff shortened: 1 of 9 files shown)


wwilsman:
Now depends on #18

app/controllers/vendors_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+      count: params[:count] || 25,
+      offset: params[:offset] || 1
+    )
+
+    # Make the RM-API request and parse the response body
+    response = make_request(uri)
+    data = JSON.parse(response.body)
+
+    # Return the list of vendors
+    if response.code === '200'
```
sadatay: do we want to check for `200` specifically here, or would the more idiomatic `response.ok?` be sufficient?

app/controllers/vendors_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+    uri.query = URI.encode_www_form(
+      search: params[:q],
+      orderby: params[:orderby] || params[:q] ? 'relevance' : 'vendorname',
+      count: params[:count] || 25,
+      offset: params[:offset] || 1
+    )
+
+    # Make the RM-API request and parse the response body
+    response = make_request(uri)
+    data = JSON.parse(response.body)
```
sadatay: this is fine as-is for now (since we should let patterns emerge before adopting them and whatnot), but as we move forward i'd like to see the logic above this line extracted somewhere.  probably not worth its own gem, but how would you feel about a service object?

```ruby
ResourceManagementService.new(path: rmapi_path, params: params).vendors
```

or, thinking about the cases where we're going to have to do multiple RMAPI calls to surface all necessary data in one response, the service object could accept operations: 

```ruby
ResourceManagementService.new(
  path: rmapi_path, 
  operations: [ 
    GetPackage.new(...),  
    GetCustomerResource.new(...),  
  ]
).process
```
We can cross that bridge when we get to it, but my main point is that we should keep our API layer as thin as possible.

app/controllers/vendors_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+      vendors = data['vendors'].map do |vendor|
+        Vendor.new(vendor)
+      end
+
+      render jsonapi: vendors,
+             meta: { totalResults: data['totalResults'] },
+             status: :ok
+    else
+      # Map RM-API errors to JSON-API error objects
+      errors = data['Errors'].map { |err| { title: err['Message'] } }
```
sadatay: `Map` or `Hashugar` could add a little syntactic sugar to your `data` hash.  Even without bringing those in, i think `JSON.parse(...)` accepts some kind of flag like `symbolize_keys: true` which could at least free you up from `data['someString']['someOtherString']` and set you up with some symbols for keys.  All of that is relatively minor, but as we start tearing through more of these endpoints it's worth keeping in mind.

app/controllers/vendors_controller.rb
wwilsman: `response.ok?` gives me an undefined method error.

app/controllers/vendors_controller.rb
```diff
@@ -0,0 +1,83 @@
...
+    response = make_request(uri)
+    data = JSON.parse(response.body)
+
+    # Return the requested vendor
+    if response.code === '200'
+      vendor = Vendor.new(data)
+      render jsonapi: vendor, status: :ok
+    else
+      # Map RMAPI errors to JSON-API error objects
+      errors = data['Errors'].map { |err| { title: err['Message'] } }
```
sadatay: small as it is, this is starting to look like something that we'll be repeating enough to warrant its own lil method or something

app/controllers/vendors_controller.rb
wwilsman: I agree, as we were submitting this PR we noted how most other controllers are going to be the same, just subbing the model name and RM-API endpoints/query params

app/controllers/vendors_controller.rb
wwilsman: Went with `response.message === 'OK'`

sadatay (approved):
quick draw mcgraw on those revisions.  looks great! 🚢

PR #18, opened 2017-10-31
Use camelCase keys for JSON API resources

wwilsman:
## Purpose
CamelCases keys so that our consuming application does not have to have any special handling of said keys

## Approach
- Uses `camelize` instead of `dasherize`
- Updates tests to use new camelized keys
- Configuration update now requires `customerId` and `apiKey` be camelCased


cherewaty (approved):

sadatay (approved):
🐫

PR #19, opened 2017-10-31
Add health endpoint

wwilsman:
## Purpose
A health endpoint is useful when in a clustered/containerized environment and we need to check when the `mod-kb-ebsco` module is available and ready to go.

## Approach
Simply adds an `/admin/health` endpoint that always responds with a 200 status code and an empty body. The reason for `/admin` is simply that all other `mod-*` modules currently in the Folio ecosystem implement various `/admin` endpoints inherited from `raml-module-builder`. So `/admin/health` is used for consistency between modules.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -4,4 +4,6 @@ Rails.application.routes.draw do
     resource :status, only: [:show]
     match '/*path' => 'proxy#index', via: [:get, :post, :put, :patch, :delete]
   end
+
+  match '/admin/health' => 'health#index', via: [:get]
 end
--- a/app/controllers/health_controller.rb
+++ b/app/controllers/health_controller.rb
@@ -1,0 +1,5 @@
+class HealthController < ActionController::Base
+  def index
+    render :json => '', :status => :ok
+  end
+end
--- a/spec/requests/health_spec.rb
+++ b/spec/requests/health_spec.rb
@@ -1,0 +1,13 @@
+require 'rails_helper'
+
+RSpec.describe "Health", type: :request do
+  describe "GET /health" do
+    before do
+      get '/admin/health'
+    end
+
+    it "returns a 200 status code" do
+      expect(response).to have_http_status(200)
+    end
+  end
+end


config/routes.rb
```diff
@@ -1,5 +1,6 @@
 Rails.application.routes.draw do
   scope '/eholdings' do
+    resource :health, only: [:show]
```
elrickvm: Should this `'health` endpoint be scoped under `/eholdings` or non-scoped to just be`/health` so from within a cluster if you want to hit this module from its IP for example would it then be `34.123.121.34/eholdings/health` or should it be `34.123.121.34/health`?

wwilsman: That's a good question. From the initial conversation the example given was `/admin/health`, so I assumed it'd be scoped to `/eholdings`. However, I just search through several `mod-*` folio modules and only `mod-notes` has a `/health` endpoint at all. And it's `/admin/health`. Since `/admin` isn't defined in the raml for `mod-notes` I don't think the scoping on this endpoint matters at all.

We better ask Hongwei what he'd like to see since the health endpoint was his suggestion

wwilsman: @elrickvm after some more research, it looks like `mod-notes` and `mod-notify` reference `/admin/health` in their tests. Searching up the grapevine, `raml-module-builder` implements the `/admin` endpoints that apparently all of their raml-based modules inherit.

I'm going to adjust this to also be `/admin/health` so that it's in-line with what's expected of the other modules.

elrickvm: @wwilsman makes sense if they use it for health checking the a pod. Seems like this should be implemented at the container level in the Dockerfile or like us in the Kubernetes yaml. Reason i say that is because this doesn't really give them the granular feedback of whats actually happening with the container but lets just align with whatever they are doing for sure. Yeah we can double check with Hongwei to make sure.

wwilsman: @elrickvm Relevant part of the `raml-module-builder`'s readme: https://github.com/folio-org/raml-module-builder/#monitoring

elrickvm (approved):

PR #21, opened 2017-11-06
JSON API Titles

wwilsman:
## Purpose
Abstract RM API requests into a service and implement a JSON API endpoint for titles

## Approach
- I noticed that every controller that inherits from the application controller also needs to create a configuration object at some point. So I moved the creation of that configuration object into the application controller so it can be shared by future controllers as well.

- Created an RM API Service that allows us to make RM API requests without the hassle of including the `rmapi_base_url`, `customer_id`, and `api_key` every time.

- Everything implemented for vendors (#17), but for titles now

- `pubType` has become `publicationType` to avoid confusion with `publisher`

- `*List` attributes are now simply plural attributes (`identifiersList => identifiers`)

- `publicationType` is now the string representation instead of an integer

- `identifiers` `.type` & `.subtype` are also now strings instead of integers

- `customerResources` are no longer included by default, but can be included with the `include` query param (`?include=customerResources`). The included resources conform to the JSON API spec too, of course.

- Customer Resources now have their very own `id`! Since `vendorId`, `packageId`, and `titleId` are all already unique to their respective resources, a Customer Resource `id` is simply a composite of the others: `"<vendorId>-<packageId>-<titleId>"`.

- Customer Resources also only implement the minimum attributes needed for the front end app to render a Title with a list of Customer Resources below it.

#### Future Work:
- Add JSON API `links` to the various resources


app/controllers/titles_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  private
+
+  def rmapi
+    @rmapi ||= ::RmApiService.new(
+      base_url: rmapi_base_url,
+      customer_id: config.customer_id,
+      api_key: config.api_key
+    )
+  end
```
sadatay, line 45: any reason not to pull this out to the application controller?

sadatay:
:shipit: lgtm!

app/controllers/titles_controller.rb
wwilsman: My only reasoning is status and configuration do not need it. ~Although configuration uses the service, it uses it's own attrs instead of the config to initialize it.~ I confused the configuration model with the controller, the status and config _controllers_ do not use the service.

sadatay: ah word, works for me

sadatay (approved):

app/controllers/vendors_controller.rb
```diff
@@ -1,85 +1,44 @@
...
-    # Create and send the request
-    http.get(
-      uri.request_uri,
-      {
-        "X-Api-Key" => configuration.api_key,
-        "Content-Type" => 'application/json',
-        "Accept" => 'application/json'
-      }
+  def rmapi
+    @rmapi ||= ::RmApiService.new(
```
cowboyd, line 38: Is there a need to memoize this state?

wwilsman: Yea, the service has two attributes we need access to when a request fails `rmapi.errors` and `rmapi.response.status`. We could have the request return the response, but then we'd have to repeat the parsing of JSON everywhere

cowboyd: Rather than mutating the service, why not have the `success!` and `fail!` return `RmapiService::Response` objects. Then you get the best of both worlds.

wwilsman: That sounds much better actually. Will start on those changes

PR #24, opened 2017-11-07
Move JSON rescue block to the bottom of the configuration update method

wwilsman:
## Purpose
I noticed that after the `rescue` block in the configuration controller the method continues execution, causing a double render.

## Approach
- Move the rescue block to the bottom of the method and specifically catch `JSON:ParserError`
- Also rescue `NoMethodError` in the event that the JSON payload does not contain the proper keys
- Add test to make sure we get the proper response code

### DIFF
--- a/app/controllers/configurations_controller.rb
+++ b/app/controllers/configurations_controller.rb
@@ -4,17 +4,7 @@ class ConfigurationsController < ApplicationController
   end
 
   def update
-    begin
-      data_attributes = JSON.parse(request.body.read)['data']['attributes'] || {}
-    rescue
-      error = {
-        title: 'Invalid JSON',
-        detail: 'The provided JSON payload could not be parsed'
-      }
-
-      render jsonapi_errors: [error],
-             status: :unprocessable_entity
-    end
+    data_attributes = JSON.parse(request.body.read)['data']['attributes'] || {}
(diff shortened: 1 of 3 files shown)


elrickvm (approved):

sadatay (approved):

sadatay:
when we get more into deserialization i'm guessing we'll end up pulling that parse block out somewhere, but for now this lgtm

cowboyd (changes requested):
The problem here is that we're potentially doing a double-render in the event of an exception.

Can we not remove the `begin` / `rescue` / `end` altogether and move the `rescue` down to the bottom?

wwilsman:
@cowboyd that’s why I added the return. To prevent the double render. If we moved the rescue block to be for the entire method, it might also render the JSON error for an unrelated exception.

cowboyd:
@wwilsman We would still want to render an error though, right? In that case, could we not use a separate `rescue` block for each type of error to render?

wwilsman:
@cowboyd I forgot you can rescue specific exceptions in Ruby. Adjusted.

cowboyd:
Ruby, you're a-ok 👌

cowboyd (approved):

PR #25, opened 2017-11-08
Implement packages JSON API endpoint

wwilsman:
## Purpose
Implements a JSON API endpoint for packages and also allows you to include customer resources via `?include=customerResources`.

## Approach
- Adds a new `request_multi` method to the RM API Service which can make multiple requests and returns the responses with combined `ok?` and `errors` methods.

  ```ruby
  # requests are made in the order defined until one fails
  responses = rmapi.request_multi(
    package: [:get, '/vendors/19/packages/6581'],
    titles: [:get, '/vendors/19/packages/6581/titles?count=25']
  )

  responses.ok? 
  # => true if all responses are OK

  responses.errors
  # => errors from the last failed request

  responses.code
  # => the last response's code

  responses.package
  # => package request response

  responses.titles
  # => titles request response
  ```

- Formats package responses as JSON API

- Formats a package's `visibilityData` as always displaying "All titles in this package are hidden" rather than the RM API message

- Formats a package's `contentType` attribute

- Adds the ability to include customer resources when requesting a package (this creates another request to RM API for the related package titles)

### DIFF
--- a/app/controllers/packages_controller.rb
+++ b/app/controllers/packages_controller.rb
@@ -1,0 +1,75 @@
+class PackagesController < ApplicationController
+  def index
+    # Transform query params to what the EBSCO RM-API expects
+    query = URI.encode_www_form(
+      search: params[:q],
+      orderby: params[:orderby] || params[:q] ? 'relevance' : 'packagename',
+      count: params[:count] || 25,
+      offset: params[:offset] || 1
+    )
+
+    # Make the request for packages from the RM API
+    response = rmapi.request(:get, "packages?%{query}" % { query: query })
+
+    if response.ok?
+      render jsonapi: response.data.packagesList.map { |package| Package.new(data: package) },
+             meta: { totalResults: response.data.totalResults }
+    else
+      render jsonapi_errors: response.errors,
+             status: response.code
+    end
+  end
+
+  def show
+    # The package id is a composite of the vendor id since package
+    # resources are nested within vendors
+    vendor_id, package_id = params[:id].split('-')
+    included_resources = params[:include] ? params[:include].split(',') : []
+    pending_requests = {}
+
+    package_path = "vendors/%{vendor_id}/packages/%{package_id}" % {
+      vendor_id: vendor_id || 0,
+      package_id: package_id || 0
+    }
+    pending_requests[:package] = [:get, package_path]
+
+    if included_resources.include?('customerResources')
+      package_titles_path = "#{package_path}/titles?%{query}" % {
+        # Required RM API params
+        query: URI.encode_www_form(
+          search: '',
+          searchfield: 'titlename',
+          orderby: 'titlename',
+          count: 25,
+          offset: 1
+        )
+      }
+      pending_requests[:titles] = [:get,  package_titles_path]
+    end
+
+    # This is a 'compound' request, consisting of multiple calls
+    # to the RMAPI. `request_multi` will process each request
+    # in the pending_requests object and return the responses
+    # in a similar data structure
+    responses = rmapi.request_multi(pending_requests)
+
+    if responses.ok?
+      titles = responses.titles.data.titles rescue []
+      render jsonapi: Package.new(data: responses.package.data, titles: titles),
+             include: params[:include]
+    else
+      render jsonapi_errors: responses.errors,
+             status: responses.code
+    end
+  end
+
+  private
+
+  def rmapi
+    RmApiService.new(
+      base_url: rmapi_base_url,
+      customer_id: config.customer_id,
+      api_key: config.api_key
+    )
+  end
+end
--- a/app/services/rm_api_service.rb
+++ b/app/services/rm_api_service.rb
@@ -18,6 +18,18 @@ class RmApiService
     end
   end
 
+  def request_multi(pending)
+    responses = {}
+
+    pending.each do |key, args|
+      res = request(*args)
+      responses[key] = res
+      break unless res.ok?
+    end
+
+    Responses.new(responses)
+  end
+
   private
 
   def uri_for(path)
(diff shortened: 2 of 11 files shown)


app/services/rm_api_service.rb
```diff
@@ -18,6 +18,18 @@ def request(verb, path, body = nil)
     end
   end
 
+  def request_multi(pending)
+    responses = {}
+
+    pending.each do |key, args|
+      res = request(*args)
+      responses[key] = res
+      break unless res.ok?
```
wwilsman, line 27: I was thinking, do we want to just continue with the remaining requests anyway here? Maybe in the future we want to still include other request data that succeeded. The order they're added to `pending` becomes the order they're requested, so I wouldn't want any headaches from ordering the `pending` hash properly.

app/services/rm_api_service.rb
```diff
@@ -64,4 +76,27 @@ def errors
...
+    def ok?
+      @responses.all?(&:ok?)
+    end
+
+    def code
+      @responses.last.code
+    end
+
+    def errors
+      @responses.reduce([]) do |memo, res|
```
wwilsman: If we do want to keep the break point above in `request_multi`, this shouldn't be necessary since only the very last request will have been not OK. (same goes for the `ok?` method above)

app/services/rm_api_service.rb
```diff
@@ -18,6 +18,18 @@ def request(verb, path, body = nil)
     end
   end
 
+  def request_multi(pending)
```
cowboyd, line 21: There is no `funcadelic.rb`, but this is almost certainly a `Monoid`. Should there be? 🎸

app/controllers/packages_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+        )
+      }
+      pending_requests[:titles] = [:get,  package_titles_path]
+    end
+
+    # This is a 'compound' request, consisting of multiple calls
+    # to the RMAPI. `request_multi` will process each request
+    # in the pending_requests object and return the responses
+    # in a similar data structure
+    responses = rmapi.request_multi(pending_requests)
```
cowboyd, line 54: This is a lot of non `HTTP` logic to be in the controller. What can we move into a static method on the `Package` model?

app/services/rm_api_service.rb
cowboyd: I think we want to steer clear of that for now and treat it as transactional: complete success or failure.

app/controllers/packages_controller.rb
wwilsman: @cowboyd Hmm, the only thing I can see that might not be request logic is the RM API specific paths (and the consequential query params). Should we have those path creations in static methods of the respective models?

So `Vendor.path_for(params[:id])`, `Package.path_for(params[:id])`, `Package.path_for_titles(params[:id])`, etc?

That would clean up this method to be more like:
```ruby
def show
  included_resources = params[:include] ? params[:include].split(',') : []
  pending_requests = {}

  # Request the package
  pending_requests[:package] = [:get, Package.path_for(params[:id])]

  # Request customer resources via package titles
  if included_resources.include?('customerResources')
    pending_requests[:titles] = [:get, Package.path_for_titles(params[:id])]
  end

  # Make both requests
  response = rmapi.request_multi(pending_requests)

  if responses.ok?
    titles = responses.titles.data.titles rescue []
    render jsonapi: Package.new(data: responses.package.data, titles: titles),
           include: params[:include]
  else
    render jsonapi_errors: responses.errors,
           status: responses.code
end
```

app/services/rm_api_service.rb
wwilsman: Alright we'll keep this, but the logic in `ok?` and `errors` for `Responses` has been updated to only return the _last_ response data since that would be the only failed one in the set given to initialize.

app/controllers/packages_controller.rb
cowboyd: When I say HTTP, I mean the HTTP interface of the rails app, Actually making the requests to underlying layers belongs more raturally to the models. E.g.

```ruby
package = Package.find(params[:id], included_resources)
if package.errors?
  render jsonapi_errors: package.errors
else
  render jsonapi: package
end
```

app/services/rm_api_service.rb
cowboyd: Ah, I see, still transactional, but wait for all requests? I think that's fine either way. What we want to avoid is, at the endpoint level any idea of a "partial success"

app/controllers/packages_controller.rb
wwilsman: So the models should be making the requests then? This is different than what we've been doing so far for the RMAPI stuff. Shall we just merge this then and I can open another PR adjusting all of the controllers for vendors, titles, and packages?

That way this PR is just implementing mostly the same pattern we've been using. And another PR can be concerned with the cleanup of them all.

sadatay: I'm with @wwilsman on this one - i think you're on to something here with moving this logic out of the controller, but where it ends up and what form it takes are questions worthy of their own whiteboard sessions.  for now i'm comfortable with this going in in its current state so that UI side of the equation can keep moving today

app/services/rm_api_service.rb
sadatay: starchild

_if you hear any noise, it's just me and monoids_

app/controllers/packages_controller.rb
cowboyd: > So the models should be making the requests then? This is different than what we've been doing so far for the RMAPI stuff. 

This feels more consistent with the way rails apps are typically structured. When you have an active record, it is responsible for making the requests to query and deserialize results over the network, just from a postgres server instead of an RMAPI server.

>Shall we just merge this then and I can open another PR adjusting all of the controllers for vendors, titles, and packages?

Sounds good.

cowboyd (approved):

sadatay (approved):

PR #26, opened 2017-11-09
Implement JSONAPI Customer Resources

elrickvm:
## Purpose
Reworks the definition of a `customer-resource` to be closer to that of a "package title".  When a title is retrieved from the RMAPI in the context of a package, the `customerResourcesList` attribute of that title is an array with a single item.  Conceptually, the "Package Title" is a combination of fields on the title and fields on that nested "customer resource" object. 

The resource we are providing with this endpoint is also called a "Customer Resource", though this term is overloaded.  Our customer resource is a new data structure where data from the title and its nested resource are flattened, providing a object that is more representative of the data that the UI needs to render the package title details page. 

## Approach
- return a composite id that is combination of the `vendor_id`-`package_id`-`title_id` ie. ` "id"=>"22-1887786-1440285"`
- add `/eholdings/jsonapi/customer-resources/:id` endpoint
  - with the composite id: `/eholdings/jsonapi/customer-resources/22-1887786-1440285`
- return new flattened data structure that is JSONAPI compliant
<details>
<summary>Before</summary>

```
{
  "description": null,
  "edition": null,
  "isPeerReviewed": false,
  "contributorsList": [
    {
      "type": "author",
      "contributor": "Havard, Margaret"
    },
    {
      "type": "author",
      "contributor": "Tiziani, Adriana."
    }
  ],
  "titleId": 1440285,
  "titleName": "Havard's Nursing Guide to Drugs (Nursing Guide to Drugs)",
  "publisherName": "Elsevier",
  "identifiersList": [
    {
      "id": "1440285",
      "source": "AtoZ",
      "subtype": 0,
      "type": 9
    },
    {
      "id": "475765",
      "source": "ResourceIdentifier",
      "subtype": 0,
      "type": 16
    },
    {
      "id": "978-0-7295-3913-5",
      "source": "ResourceIdentifier",
      "subtype": 1,
      "type": 6
    },
    {
      "id": "978-0-7295-7913-1",
      "source": "ResourceIdentifier",
      "subtype": 2,
      "type": 8
    }
  ],
  "subjectsList": [
    {
      "type": "BISAC",
      "subject": "MEDICAL \/ Nursing \/ Pharmacology"
    }
  ],
  "isTitleCustom": false,
  "pubType": "Book",
  "customerResourcesList": [
    {
      "titleId": 1440285,
      "packageId": 1887786,
      "packageName": "ProQuest Ebook Central",
      "isPackageCustom": false,
      "vendorId": 22,
[...]

PR #28, opened 2017-11-20
Use ActiveRecord-esque Models for RMAPI Resources

sadatay:
The intent of this work is to standardize all our models according to the conventions of ActiveRecord/ActiveModel.  To do this I created a base class for models that pull ActiveModel behavior in, as well as request brokerage via [flexirest](https://github.com/flexirest/flexirest) which buys us a clean DSL and ETag caching.

As much as possible has been pushed to the serialization layers.  Models now behave like AR models, i.e. `Package.find(params[:id])`, `Package.all`, etc.  All inclusion relationships have been wired up and tested.

**Included Resources / Relationships**
- Vendor Search / Detail `/eholdings/jsonapi/vendors/(:vendorId)`
  - `has_many` Packages `?include=packages`
- Package Search / Detail `/eholdings/jsonapi/packages/(:vendorId-:packageId)`
  - `has_one` Vendor `?include=vendor`
  - `has_many` Customer Resources `?include=customerResources`
- Title Search / Detail `/eholdings/jsonapi/titles/(:titleId)`
  - `has_many` Customer Resources `?include=customerResources`
- Customer Resource Detail `/eholdings/jsonapi/packages/(:vendorId-:packageId-:titleId)`
  - `has_one` Vendor `?include=vendor`
  - `has_one` Title `?include=title`
  - `has_one` Package `?include=package`

**Added 11/22**
- Given the issues @cherewaty and myself ran into yesterday with RMAPI sandbox, we converted the whole test suite to run against production Okapi and production RMAPI by default.  
  - We pulled variables that are common to all request tests (`TEST_OKAPI_TOKEN`, `headers`, etc.) out into a shared context helper and exposed the url and tenant as `TEST_OKAPI_{TENANT, URL}`, which will come in handy when this is passed off to Folio org.  
  - VCR cassettes were re-recorded on production data and specs were tweaked to account for new data.
- With a passing test suite in hand we were able to implement the `around_action` thing with confidence and clean up all the controllers.

**Added 11/27**
- New work from @cowboyd moves RMAPI configuration to the model's instance level, and sidesteps some limitations of Felxirest.  

Closes #26


ModuleDescriptor.json
```diff
@@ -26,6 +26,10 @@
           "methods": [ "GET" ],
           "pathPattern": "/eholdings/jsonapi/titles*"
         },
+        {
+          "methods": [ "GET" ],
+          "pathPattern": "/eholdings/jsonapi/customer-resources*"
```
sadatay, line 31: if the resource type is `customerResources` maybe we should use camelCase here too, as much as it pains me.

app/controllers/customer_resources_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+    begin
+      vendor_id, package_id, title_id = params[:id].split('-')
+      customer_resource = CustomerResource.find(
+        vendor_id: vendor_id,
+        package_id: package_id,
+        title_id: title_id
+      )
+      render jsonapi: customer_resource,
+             include: params[:include]
+    rescue Flexirest::HTTPClientException, Flexirest::HTTPServerException, Flexirest::HTTPNotFoundClientException => e
```
sadatay: this begin/rescue is cumbersome and pops up all over the place.  best thing i could think of is to implement methods in the superclass that wrap a `yield`, but that's a little heavy so I didn't want to do it without fielding other opinions.

sadatay:
the `configuration` and `status` endpoints haven't been moved over to this new paradigm, so I haven't destroyed `RmApiService` entirely yet.

wwilsman:
@sadatay does this include fetching relationship via:  
`/eholdings/jsonapi/packages/:id/customer-resources`

This also furthers the question about camelcasing our endpoints because the actual relationship key is `customerResources`.

I ask because in the UI when we select a package we refresh the customer-resources below it. I suppose if the relationship links aren't working, we could just refresh the package in the background with `?include=customerResources`. But then when we get to paginating the relationships we would probably want something like the link above.

sadatay:
@wwilsman this iteration didn't go so far as to support that style of endpoint - just the `include` query param.  it probably wouldn't be too much additional work to just wire it up in `routes.rb` since the actual business logic is (i think) all already present.  so do we want to do that in this PR, or create another small PR that adds those endpoints and additionally exposes them as `link` properties in the `relationships` definition?

either way, i think that having these endpoints definitely points us in the direction of using camelCase in the URI to make the resource naming consistent.  it feels a little weird, but not as weird as having `customer-resources` only appear in the endpoint and nowhere else - especially when we're now talking about multiple endpoints

wwilsman:
Another PR is fine with me since refactoring the models and adding those links are kinda two separate concerns.

Another reason I bring it up is because RM API does have these URLs for querying customer-resources of a specific package. So the frontend has always queried the additional resources this way

sadatay:
word.  that raises another question - do we want endpoints like that for every relationship that has been fleshed out here?  it sounds like you're primarily concerned with packages->customerResources.  couldn't hurt to do them all, but things like `/eholdings/jsonapi/packages/(:vendorId-:packageId-:titleId)/vendor` feel sort of ancillary.  from where i'm sitting it seems like we really only care about the `has_many` associations, but let me know how you feel about it

wwilsman:
Yea, we definitely only care about the `has_many` associations right now because those are the only relationships we're actively using.

Gemfile.lock
```diff
@@ -63,7 +63,14 @@ GEM
       dotenv (= 2.2.1)
       railties (>= 3.2, < 5.2)
     erubi (1.6.1)
+    faraday (0.13.1)
+      multipart-post (>= 1.2, < 3)
     ffi (1.9.18)
+    flexirest (1.5.3)
+      activesupport
+      crack
```
cowboyd, line 71: I'm amazed that this gem is still around.

ModuleDescriptor.json
cowboyd: Nooooooooooooooooo.

app/controllers/customer_resources_controller.rb
cowboyd: You can use an `around_action` for just this purpose.

Also every method gets it's own implicit `begin`/`end` block, so you can just have the `rescue` alone.

cowboyd (commented):
Flexirest is pretty amazing.... However, rather than explicitly configuring the global flexirest resources with each request, let's try subclassing them like so:

```ruby
def customer_resources
  Class.new(CustomerResource).tap do |cls|
    cls.send(:define_method, :config) { config }
  end
end
```

cowboyd (approved):

PR #29, opened 2017-11-29
JSONAPI Customer Resource Updates

sadatay:
This seems to be functional, but could use a few more tweaks and tests.  I'll circle back with it tomorrow.

image


app/controllers/customer_resources_controller.rb
```diff
@@ -1,11 +1,25 @@
 class CustomerResourcesController < ApplicationController
   def show
     @customer_resource = customer_resources.find customer_resource_id
+
     render jsonapi: @customer_resource, include: params[:include]
   end
 
+  def update
+    @customer_resource = customer_resources.find customer_resource_id
+    customer_resources.update(
```
cowboyd: Maybe put this as an instance method?

```ruby
def update(params)
  self.class.update(with_params params)
end
```

cowboyd (commented):
Using a separate cassette for GET and PUT seems fine by me.

cowboyd (approved):

PR #30, opened 2017-12-05
Update packages with JSON API PUTs

cherewaty:
The sequel to https://github.com/thefrontside/mod-kb-ebsco/pull/29

Allows editing of `isSelected`, `isHidden`, and `customCoverage` for packages.

### DIFF
--- a/app/controllers/packages_controller.rb
+++ b/app/controllers/packages_controller.rb
@@ -1,4 +1,6 @@
 class PackagesController < ApplicationController
+  deserializable_resource :package, only: :update,
+                          class: DeserializablePackage
 
   def index
     @packages = packages.all(q: params[:q])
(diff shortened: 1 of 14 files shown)


app/controllers/packages_controller.rb
```diff
@@ -1,4 +1,8 @@
 class PackagesController < ApplicationController
+  attr_accessor :package
```
cowboyd: Is this accessor ever used?

app/controllers/packages_controller.rb
```diff
@@ -11,14 +15,31 @@ def show
     render jsonapi: @package, include: params[:include]
   end
 
+  def update
+    @package = packages.find package_id
+    @package.update package_params
```
cowboyd, line 18: The "AR" way to do this would be:

```ruby
@package = packages.update package_id, package_params
render status: :no_content
```
I know that's not the way that customer resources does it, but worth considering changing them both.... or maybe with the next update.

app/controllers/packages_controller.rb
cherewaty: Nope, removed.

app/controllers/packages_controller.rb
cherewaty: Let's revisit that later.

cowboyd (approved):

PR #31, opened 2017-12-05
Make sure contributors and subjects are always arrays

cherewaty:
The EBSCO RM API usually returns an array for `subjectsList` and `contributorsList`, but not always.

In the response from RM API's `/vendors/19/packages/6581/titles?count=25&offset=1&orderby=titlename&search=&searchfield=titlename`, one of customer resources has a `subjectsList` of `null`.

This will ensure `subjects` and `contributors` attributes on titles and customer resources coming back from `mod-kb-ebsco` always return an array (but an empty one if RM API returned `null`).


cowboyd (approved):
🙌

PR #32, opened 2017-12-07
Adjust isHidden deserialization

cherewaty:
## Background
- RM API gives us a package's `isHidden` attribute inside of a `visibilityData` object.
- To edit, RM API expects `isHidden` to be a top-level key in a PUT, not inside of `visibilityData`.
- The de/serialization taking place in https://github.com/thefrontside/mod-kb-ebsco/pull/30 expected `mod-kb-ebsco` consumers to PUT `isHidden` on the top level of attributes, not within `visibilityData`.

## Approach
This makes `mod-kb-ebsco`'s GETs and PUTs symmetrical, with `isHidden` always inside of `visibilityData`.

We may want to flatten it out at some point, with `isHidden` and `isHiddenReason` as un-nested top-level attributes, but this will work with `ui-eholdings` as currently built.

## Bonus
I documented a few of RM API's quirkier behaviors in the packages spec, but didn't attempt to solve for them better yet. When a package is not selected, its custom coverage and hidden status aren't editable, but RM API still returns a 204. In that scenario, we should probably return a 500 from `mod-kb-ebsco` with a useful error message instead.

### DIFF
--- a/app/deserializable/deserializable_customer_resource.rb
+++ b/app/deserializable/deserializable_customer_resource.rb
@@ -1,12 +1,9 @@
 class DeserializableCustomerResource < JSONAPI::Deserializable::Resource
   attributes :isSelected,
-             :customEmbargoPeriod
+             :customEmbargoPeriod,
+             :visibilityData
 
   attribute :customCoverages do |value|
     { customCoverageList: value }
-  end
-
-  attribute :isHidden do |value|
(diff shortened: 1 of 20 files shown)


cherewaty:
Commit added with the same `isHidden` deserialization changes for customer resources (but without the testing reorganization I did for packages).

cowboyd (approved):

PR #35, opened 2017-12-11
Return composite package id for customer resource

cherewaty:
## Purpose
With `ui-eholdings` hitting the new JSON API endpoints, this package link was broken.
screen shot 2017-12-11 at 8 14 51 am

## Approach
Serialize a customer resource's `packageId` to be the composite format that includes vendor ID.


sadatay (approved):
👍 right on

PR #37, opened 2017-12-12
Revert manual deserialization

cherewaty:
The refactor in https://github.com/thefrontside/mod-kb-ebsco/pull/34 produced the exact same behavior, but less elegantly than before.


sadatay (approved):
👍 danke schoen

PR #38, opened 2017-12-13
Return response bodies from PUTs

cherewaty:
## Purpose
The EBSCO RM API has business logic that it doesn't always surface in API responses. Example: when unselecting a package, `customCoverage` gets removed. In that situation, RM API returns a `204 No Content`.

## Approach
We can re-fetch packages and customer resources from the RM API immediately after `PUT`ting changes to them. That means `mod-kb-ebsco` is making twice as many calls out, but consuming apps will receive any unintended side-effects to attributes in `mod-kb-ebsco`'s response to a `PUT`.

## What this doesn't solve
Customer resources and packages are pretty tightly coupled. Example: when unselecting a package, all of its customer resources become unselected. Relationship side-effects were not addressed here. We should evaluate the strategy for that problem after pagination has been built.


cherewaty:
Best kind of PR 😀 

screen shot 2017-12-12 at 4 03 02 pm

wwilsman:
RE: Relationship side-effects

Currently, the frontend simply refreshes the current route relationships when an update to the record happens. For example: when you select a package, a second request for it's customer-resources is made after a response is received for the select request

app/controllers/customer_resources_controller.rb
```diff
@@ -14,7 +14,9 @@ def show
   def update
     @customer_resource.update customer_resource_params
 
-    render status: :no_content
+    # re-fetch from RM API to surface side-effects
+    @customer_resource = customer_resources.find customer_resource_id
```
cowboyd: Should this be handled at the model layer? e.g.

```ruby
@customer_resource = customer_resources.update customer_resource_id, customer_resourc_params
```

cherewaty: Re-fetch moved to model layer.

app/controllers/packages_controller.rb
```diff
@@ -16,8 +16,7 @@ def show
   def update
     @package = packages.find package_id
     @package.update package_params
```
cowboyd, line 18: AR allows you to update a record from the collection. Is that hard to do here?

```ruby
@package = packages.update package_id, package_params
render jsonapi: @package
```

app/controllers/customer_resources_controller.rb
```diff
@@ -13,8 +13,7 @@ def show
 
   def update
     @customer_resource.update customer_resource_params
```
cowboyd, line 15: In AR, a collection can update a single record via the id. This seems like less magic, and more in keeping with convention. Is there anything preventing this here?

```ruby
@resource = customer_resources.update customer_resource_id, customer_resource_params
render jsonapi: @resource
```

cowboyd (commented):
Minor suggested tweaks to the api.

app/controllers/customer_resources_controller.rb
cherewaty: It's not how flexirest works out of the box: https://github.com/flexirest/flexirest/blob/a4b89653b672cdffb3f084f37bbd81ee77c09afa/Ruby-on-Rails-Integration.md#controller

Not sure it's worth the extra steps to get that syntax.

sadatay (approved):
👍

PR #39, opened 2017-12-15
Add Transfer-Encoding: chunked middleware

cherewaty:
Okapi sends proxied requests with a `Transfer-Encoding: chunked` header. `mod-kb-ebsco` should be able to handle chunked requests.

Puma appears to have fixed the issues mentioned in the middleware usage notes: https://github.com/puma/puma/commit/5ec232d1d47283195275acccf8d53010b22416f2


cowboyd:
Do we use Puma?

cherewaty:
I believe so?

cowboyd (approved):

PR #41, opened 2017-12-15
Accept page param for pagination

cherewaty:
Searching for vendors, packages, or titles now accept a `?page=` param that gets mapped to the EBSCO RM API's `?offset=` (both 1-indexed).

This might also work for paginating included resources, but I didn't put any test coverage around that, so no guarantees.


app/models/customer_resource.rb
```diff
@@ -22,7 +22,8 @@ class CustomerResource < RmApiResource
       request.get_params[:searchfield] = 'titlename'
       request.get_params[:orderby] = 'titlename'
       request.get_params[:count] ||= 25
-      request.get_params[:offset] ||= 1
+      request.get_params[:offset] = (request.get_params[:page] ? request.get_params[:page] : 1)
```
wwilsman: Can this be an `||` operator? `request.get_params[:page] || 1`

Same goes for `package` and `title` too

spec/requests/vendors_spec.rb
```diff
@@ -5,16 +5,33 @@
...
+    describe "with pagination" do
+      before do
+        VCR.use_cassette("search-vendors-page2") do
+          get '/eholdings/jsonapi/vendors/?q=e&page=2', headers: okapi_headers
+        end
+      end
+
+      let!(:json2) { Map JSON.parse response.body }
+
+      it "gets a different list of resources" do
```
wwilsman, line 29: I don't think this test should be passing. `page => offset` transform doesn't seem to have been implemented for the `vendor` model. 🤔

cherewaty: Yup, I should've been using `not_to eql` instead of `not_to equal`. (https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/equality-matchers)

Fixed.

wwilsman (approved):

PR #44, opened 2017-12-18
Access related resources with compound endpoints

sadatay:
This allows accessing related resources with JSONAPI-compliant endpoints.  For instance:

- `GET /jsonapi/packages/19-2401?include=customerResources` will return the package and the related customer resources.
- `GET /jsonapi/packages/19-2401/customer-resources` will return only the related customer resources.

Our frontend data layer depends on endpoints of this format in order to properly handle the reloading of related data when resources are updated.

**Caveat:** Due to a discrepancy in response fields found in RMAPI, there is a minor bug where certain fields are omitted from the `packages/(package-id)/customer-resources` requests.  See [here](https://developer.ebsco.com/rmapireference#titlemanaged) and [here](https://developer.ebsco.com/rmapireference#titlepackageid) for more details.  For now we've agreed to live wiith this, and the RMAPI team is taking a look at it.


app/controllers/packages_controller.rb
```diff
@@ -1,6 +1,8 @@
 class PackagesController < ApplicationController
-  deserializable_resource :package, only: :update,
-                          class: DeserializablePackage
```
cherewaty, line 3: Good simplification, I thought that might be possible but hadn't tried it.

cherewaty (approved):
👍 Needs a rebase, looks great.

PR #45, opened 2017-12-19
Return failure code when trying to customize a deselected package

sadatay:
This addresses the `TODO`s in the packages spec.  While our UI will not attempt to customize deselected resources, a manual API request to our service could result in a confusing state: a success code with a response body not matching the desired updates.  This rejects any attempted customization to deselected packages with a 422 (Unprocessable Entity) status code, and a response body with the relevant error(s).

### DIFF
--- a/app/controllers/packages_controller.rb
+++ b/app/controllers/packages_controller.rb
@@ -1,3 +1,4 @@
+
 class PackagesController < ApplicationController
 
   before_action :set_package, only: [:show, :update, :customer_resources]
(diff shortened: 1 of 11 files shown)


sadatay:
One thing I noticed about the VCR tests is that they must be executed in a specific order - i.e. the state set by the previous test is carried over into the next test.  It might be a good idea in the future do so some kind of per-test setup and teardown.  Also, gonna do some hacking tonight on a one-liner to "rewind" all cassettes for a given spec file, because doing that manually is a PITA.

cherewaty (approved):

PR #46, opened 2017-12-20
Remove /jsonapi namespace

cherewaty:
Now that `ui-eholdings` exclusively uses the new JSON API endpoints, we no longer need them sandboxed under `/jsonapi/`.

The older endpoints that simply proxied requests to the EBSCO RM API can now be found under `/proxy/`.


ModuleDescriptor.json
```diff
@@ -16,19 +16,19 @@
...
-          "pathPattern": "/eholdings/jsonapi/titles*"
+          "pathPattern": "/eholdings/titles*"
         },
         {
           "methods": [ "GET", "PUT" ],
-          "pathPattern": "/eholdings/jsonapi/customer-resources*"
+          "pathPattern": "/eholdings/customer-resources*"
         },
         {
           "methods": [ "GET", "POST", "PUT", "PATCH", "DELETE" ],
```
sadatay, line 34: the routes below these should probably be adjusted for the new proxy endpoints, if we keep them at all

sadatay (changes requested):
we'll also want to make sure we don't merge this until the corresponding UI work has been merged - either that or we do some kind of cutover where we keep both the `/jsonapi` prefixed endpoints and the new endpoints around until the UI is adjusted

ModuleDescriptor.json
cherewaty: Routes in module descriptor updated to include `proxy`.

sadatay (approved):
:shipit:

PR #48, opened 2017-12-20
Move access for raw EBSCO RM API endpoints

cherewaty:
`proxy` is a loaded word in the FOLIO ecosystem, and we weren't using it in the same way as Okapi expects.

Now the raw access to the EBSCO RM API endpoints will be under `/ebsco-rmapi` instead of `/eholdings/proxy`.


sadatay (approved):
⛵️

PR #65, opened 2018-01-16
Modify visibility data reason that we get from RM API as needed by ui-eholdings for display

sduvvuri-ebsco:


spec/requests/customer_resources_spec.rb
```diff
@@ -72,6 +72,11 @@
       it 'has reason ' do
         expect(attributes.visibilityData).to have_key(:reason)
       end
+      it 'has reason empty' do
+        expect(attributes.visibilityData.reason).to(
+          eq('')
+        )
```
sadatay: while this is valid, it'd be slightly more idiomatic ruby to drop the parens from `to`, like so:
```
expect(attributes.visibilityData.reason).to eq('')
```

PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,14 @@
...
+_How does this change fulfill the purpose?_
+
+#### TODOS and Open Questions
+- [ ] Use GitHub checklists. When solved, check the box and explain the answer.
+
+## Learning
+_Describe the research stage. Add links to blog posts, patterns, libraries or addons used to solve this problem._
+
+## Screenshots
+_Let's see those sweet visuals!_
```
sadatay, line 14: i'm unsure why this file is in this PR, since it should be in master already as of a while ago, right?  did we lose it somewhere along the way?  maybe i'm confusing this with ui-eholdings...

spec/requests/customer_resources_spec.rb
```diff
@@ -515,4 +520,46 @@
...
+    let!(:json) { Map JSON.parse response.body }
+    let!(:visibility) { json.data.attributes.visibilityData }
+
+    it 'gets a successful response' do
+      expect(response).to have_http_status(200)
+    end
+    it 'has reason hidden by ep' do
+      expect(visibility.reason).to(
+        eq('Set by System')
+      )
```
sadatay: same here, you can probably knock this down to one line and take out the parens

sadatay:
other than some minor, stylistic stuff this is looking good to me 👍

spec/requests/customer_resources_spec.rb
sduvvuri-ebsco: will take care of this, thanks

PULL_REQUEST_TEMPLATE.md
sduvvuri-ebsco: Its not in master as of this PR, yes, it is in ui-eholdings but missing from here.

sadatay (approved):

PR #67, opened 2018-01-18
Implement Package Level Hiding reason Change

carolegodfrey:
## Purpose
Update hidden reason returned as package level from rm api.  When package is hidden and reason is 'Hidden by EP' set to 'Set By System' otherwise it is set to empty.  Currently message returned is 'All titles in this package are hidden' and there is no way to distinguish if it is hidden by ep or a customer.
Related Jira ticket:
https://issues.folio.org/browse/UIEH-52

## Approach
Used same approach as was done for package title level hide reason. Added VCR cassettes for testing response with Hidden by EP and Hidden by Customer

## Learning
Ruby setup on linux https://gorails.com/setup/ubuntu/16.04
https://github.com/vcr/vcr

### DIFF
--- a/app/serializable/serializable_package.rb
+++ b/app/serializable/serializable_package.rb
@@ -22,11 +22,11 @@ class SerializablePackage < SerializableResource
   attribute :visibilityData do
     visibility = @object.visibilityData
 
-    if visibility['isHidden']
-      { isHidden: true, reason: 'All titles in this package are hidden' }
-    else
-      visibility
+    if visibility[:isHidden]
+      visibility[:reason] =
+        visibility[:reason] == 'Hidden by EP' ? 'Set by System' : ''
     end
+    visibility
   end
 
   attribute :contentType do
--- a/spec/fixtures/vcr_cassettes/get-package-reason-hidden-by-customer.yml
+++ b/spec/fixtures/vcr_cassettes/get-package-reason-hidden-by-customer.yml
@@ -1,0 +1,175 @@
+---
+http_interactions:
+- request:
+    method: get
+    uri: https://okapi.frontside.io/configurations/entries?query=module=KB_EBSCO
+    body:
+      encoding: US-ASCII
+      string: ''
+    headers:
+      Content-Type:
+      - application/json
+      Accept:
+      - application/json
+      X-Okapi-Tenant:
+      - fs
+      X-Okapi-Token:
+      - TEST_OKAPI_TOKEN
+      Accept-Encoding:
+      - gzip;q=1.0,deflate;q=0.6,identity;q=0.3
+      User-Agent:
+      - Ruby
+  response:
+    status:
+      code: 200
+      message: OK
+    headers:
+      Server:
+      - nginx/1.13.5
+      Date:
+      - Wed, 17 Jan 2018 23:32:06 GMT
+      Content-Type:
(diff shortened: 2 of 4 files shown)


app/serializable/serializable_package.rb
```diff
@@ -23,10 +23,10 @@ class SerializablePackage < SerializableResource
     visibility = @object.visibilityData
 
     if visibility['isHidden']
-      { isHidden: true, reason: 'All titles in this package are hidden' }
-    else
-      visibility
+      visibility['reason'] =
+        visibility['reason'] == 'Hidden by EP' ? 'Set by System' : ''
```
sduvvuri-ebsco: At package level, if we are modifying the message to "Set by System" or '' , then I do not think we will know at customer resource level, whether all titles in a package are hidden. We need that information to display the message "All titles in this package are hidden" and to make the toggle switch read only in case all titles are hidden at the customer resource level.

sduvvuri-ebsco (commented):
At package level, if we are modifying the message to "Set by System" or '' , then I do not think we will know at customer resource level, whether all titles in a package are hidden. We need that information to display the message "All titles in this package are hidden" and to make the toggle switch read only in case all titles are hidden at the customer resource level.

app/serializable/serializable_package.rb
carolegodfrey: I believe you can just use the isHidden boolean indicator to know that the titles are all hidden in a package. Logic does not care if it is 'Hidden by EP' or 'Hidden by Customer' in this case.

sduvvuri-ebsco: true..shall I just handle this in front-end?

carolegodfrey: hmm - I believe so - but good point since required message is different depending upon whether you are viewing package in context of Package itself or one of it's Package-Titles

sadatay: i agree with carole here - ~~if we aren't already handling this in the UI that's where it should be handled~~ - looks like we weren't, so we should be.  this may require some fidgeting in ui-eholdings to account for, but i believe you have upcoming work in that repo as well right?  we can definitely account for it there.

sadatay:
this looks good to me so long as we make sure to handle the messaging in the UI as discussed above.  also, this is an exemplary PR description and thank you for squashing the commits ✨

app/serializable/serializable_package.rb
sadatay: also, one minor nit here which also applied to @sduvvuri-ebsco 's most recent PR but I didn't catch it then - i _think_ at this point in the execution these are hashes with indifferent access - meaning you may be able to get away with symbol keys (`visibility[:reason]`) instead of string literal keys (`visibility['reason']`).  if that is true, we should change these to use the symbol notation.  it is more [efficient](https://stackoverflow.com/questions/8189416/why-use-symbols-as-hash-keys-in-ruby) and more idiomatic.

carolegodfrey: Thanks for suggestion @sadatay - worked great - did a git commit --amend to make this change. Hopefully ok.  On added message changes - @sduvvuri-ebsco is currently working on changes in this area

sadatay (approved):

sduvvuri-ebsco (approved):

PR #69, opened 2018-01-25
Include filter param for title search

wwilsman:
## Purpose
Need a way to filter title search results by publication type.

## Approach
RMAPI uses `resourcetype` as the filtering option. Since we have a JSON API compliant interface, we need to transform the `filter` param to this `resourcetype` param.

Since a new query param is included with title search requests, the `search-titles*` cassettes have been re-recorded.


app/controllers/titles_controller.rb
```diff
@@ -4,7 +4,12 @@ class TitlesController < ApplicationController
   before_action :set_title, only: %i[show customer_resources]
 
   def index
-    @titles = titles.all(q: params[:q], page: params[:page])
+    @titles = titles.all(
+      q: params[:q],
+      page: params[:page],
+      filter: params[:filter]
```
sadatay, line 10: i wonder if there's a neater way to do this, like with ActionController params or something.  just a thought, this is fine with me as-is.

cherewaty (approved):

PR #70, opened 2018-01-25
Add providers route, return providerId and providerName in response

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-124, we want to be able to route requests containing "providers", get the corresponding vendor information from RM API and return a response containing "providerId" and "providerName" along with "vendorId" and "vendorName" for ui-eholdings to pick up.

## Approach
- Add providers to routes
- Implement a providers controller, a serilalizable and a provider model that routes requests to RM API vendors and gets the response. 
- Modify serializables of customer resource and package to provide providerId = vendorId and providerName = vendorName along with vendorId and vendorName 
- Add associated unit tests

#### TODOS and Open Questions
- [ ] Eventually, when we reach a point where we can eliminate *vendors* route, we should clean up code to get of vendors controller, serializable and model.

## Learning
http://guides.rubyonrails.org/v3.2.8/routing.html

## Screenshots
mod-kb-ebsco

### DIFF
--- a/ModuleDescriptor.json
+++ b/ModuleDescriptor.json
@@ -19,6 +19,10 @@
           "pathPattern": "/eholdings/vendors*"
         },
         {
+          "methods": [ "GET" ],
+          "pathPattern": "/eholdings/providers*"
+        },
+        {
           "methods": [ "GET", "PUT" ],
           "pathPattern": "/eholdings/packages*"
         },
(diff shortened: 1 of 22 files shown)


carolegodfrey (commented):
Looks great!  Wonder if we also should have a test in packages_spec.rb (getting a package with included provider) and  customer_resources_spec (getting a customer resource with included provider) since those relationships added?

sduvvuri-ebsco:
Sure, makes sense. Thanks for your feedback Carole. Added those unit tests as well.

carolegodfrey (approved):
Changes look good to me as to what is needed for the ui-eholdings vendor to provider mapping

cherewaty (approved):

PR #72, opened 2018-02-08
Allow additional filters for titles

wwilsman:
## Purpose
There needs to be a way to filter titles by selection status in addition to the publication type.

## Approach
JSON API specs say `filter` can be of the form `filter[attr]`. So it makes sense that we can implement the above two filters as `filter[selected]` and `filter[type]`. 

The current `filter=type` syntax is not used in eholdings, so moving this to be `filter[type]` should only effect [ui-eholdings PR #192](https://github.com/folio-org/ui-eholdings/pull/192)

I've also added a bad request error if `filter` is used without specifying an attribute. Since we use `ActionController::API`, we needed to add the ability for our application to handle `BadRequest` errors.

### TODO and Open Questions:
- [x] I simply added a `catch_exceptions` method for `around_action` to catch the `BadRequest` error. I think `ActionController::Base` has this ability out-of-the-box by combining `ActionController::API` with other sub-modules. But I'm not sure which sub-module it is that provides this.

  After some research, I'm not sure if this will even format the errors as JSON API errors. So I'll merge for now, but this can be changed in another PR if there's a better way.


spec/requests/titles_spec.rb
```diff
@@ -36,22 +35,56 @@
...
+        end
+
+        let!(:json_f2) { Map JSON.parse response.body }
+
+        it 'gets a list of unselected book resources' do
+          expect(response).to have_http_status(200)
+          expect(json_f2.data.length).to equal(25)
+          expect(json_f2.meta.totalResults).to equal(1278)
+          expect(json_f2.data.first.attributes.publicationType).to eql('Book')
+          expect(json_f2.data.first.id).not_to eql(json_f.data.first.id)
```
wwilsman, line 70: So there is no way to really test that these are selected or not because titles do not have an `isSelected` attribute. But their related customer-resources do, and _this_ is what this filter represents. For example, there are 1296 results for `?q=news&filter[type]=book`, but 1278 are "selected" and ~1230 or so are "not selected". Because a title can contain both selected and not selected customer-resources.

app/models/title.rb
```diff
@@ -7,9 +7,22 @@ class Title < RmApiResource
...
+        raise ActionController::BadRequest, 'Invalid filter parameter'
+      end
+
+      if filters[:selected] == 'true'
+        request.get_params[:selection] = 'selected'
+      elsif filters[:selected] == 'false'
+        request.get_params[:selection] = 'notselected'
+      elsif filters[:selected] == 'ebsco'
+        request.get_params[:selection] = 'orderedthroughebsco'
+      end
```
carolegodfrey: Looks good - I'm adjusting my code based on some changes here as well. Wondering if we should add a default case for selection = 'all' if it is not specified?

wwilsman: RM API defaults to all, but you're right, it's probably better to be explicit here.

carolegodfrey (approved):

PR #73, opened 2018-02-08
Implement Targeted Search Filters

carolegodfrey:
## Purpose
Updates are required to support search fields for title searches. In terms of rmapi requests - we specify both search (the search query) and searchfield  (titlename, publisher, isxn or subject) as parameters. Update mod-kb-ebsco title search to use **filter** hash parameter to support a syntax such as the following for the 4 search fields.

- _/eholdings/titles/?filter[name]=ebsco_
- _/eholdings/titles/?filter[publisher]=random_
- _/eholdings/titles/?filter[isxn]=1362-3613_
- _/eholdings/titles/?filter[subject]=history_

This syntax modifies the previous syntax which used a q parameter, such as
 - _/eholdings/titles/?q=ebsco_

These changes are server side changes required for 
https://issues.folio.org/browse/UIEH-120

## Approach

- Updated processing of query parameters in title.rb to handle hash filter values and properly specify as rmapi parameters
- Addedchecks for conflicting filter parameters 
- Added tests

#### TODOS and Open Questions
[x] Need to rebase once changes from related PR are merged https://github.com/folio-org/mod-kb-ebsco/pull/72 as used the same error handling and strategy for extracting filter values
[x] Maintain backward compatibility to honor q parameter until such time it is no longer used by ui-eholdings clients
[ ] Follow on updates to Providers and Packages endpoints to support filter[] syntax for queries for consistency

## Learning
http://jsonapi.org/format/#query-parameters
https://developer.ebsco.com/reference/rmref#get-custid-titles
http://api.rubyonrails.org/classes/ActionController/Parameters.html
https://stackoverflow.com/questions/34949505/rails-5-unable-to-retrieve-hash-values-from-parameter


app/models/title.rb
```diff
@@ -13,6 +13,39 @@ class Title < RmApiResource
...
+      publisher = filters[:publisher]
+      query = request.get_params.delete(:q)
+
+      if query && querykeys.size == 1
+        raise ActionController::BadRequest, 'Conflicting query parameters'
+      end
+
+      if query
+        request.get_params[:search] = query
+        request.get_params[:searchfield] ||= 'titlename'
```
wwilsman: I don't think we need `||=` in these cases. This implies that we will allow a `?searchfield` param to supersede any of these rules.

carolegodfrey: Thanks for catching that - made that adjustment

spec/requests/titles_spec.rb
```diff
@@ -86,6 +86,119 @@
...
+        let!(:json_t2) { Map JSON.parse response.body }
+        it 'gets a different list of resources' do
+          expect(response).to have_http_status(200)
+          expect(json_t2.data.length).to equal(25)
+          expect(json_t2.meta.totalResults).to equal(61)
+          expect(json_t2.data.first.id).not_to eql(json_t.data.first.id)
+        end
+      end
+    end
+    describe 'with search field isxn filter' do
```
wwilsman: I'm just being nitpicky now, but could we add an empty line before this like the other `describe` blocks?

spec/requests/titles_spec.rb
```diff
@@ -86,6 +86,119 @@
...
+      let!(:json_c) { Map JSON.parse response.body }
+
+      it 'returns a bad request' do
+        expect(response).to have_http_status(400)
+        expect(json_c.errors.first.title).to eql(
+          'Conflicting filter parameters'
+        )
+      end
+    end
+    describe 'with conflicting query parameters' do
```
wwilsman: Same here. :)

wwilsman (approved):
Just a small nitpick. Otherwise looks good to me.

spec/requests/titles_spec.rb
carolegodfrey: Sure makes sense and helps readability - just need to have rubocop agree :)

PR #75, opened 2018-02-15
Allow fetch and update of allowEbscoToAddTitles with RM API

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-54, users should be able to view and update "allowEbscoToAddTitles" at the package level, permission which allows knowledge base to add new titles only when a package is selected. These changes are necessary to get and put "allowEbscoToAddTitles" with RM API.
"allowEbscoToAddTitles" is a param needed by RM API and in order to keep it generic, we used "allowKbToAddTitles" as a param for mod-kb-ebsco

## Approach
This change allows us to parse allowEbscoToAddTitles" from response of RM API as well as pass it to RM API. 
Also, has associated unit tests.

## Learning
http://jsonapi.org/format/

## Screenshots
allowtitles

### DIFF
--- a/app/controllers/packages_controller.rb
+++ b/app/controllers/packages_controller.rb
@@ -56,6 +56,7 @@ class PackagesController < ApplicationController
       .require(:package)
       .permit(
         :isSelected,
+        :allowEbscoToAddTitles,
         visibilityData: [:isHidden],
         customCoverage: %i[beginCoverage endCoverage]
       )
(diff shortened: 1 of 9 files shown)


cherewaty:
`allowEbscoToAddTitles` is the RM API name, but should we make the attribute `mod-kb-ebsco` exposes more generic? `allowKbToAddTitles`?

sduvvuri-ebsco:
sure, makes sense, will make that change

sadatay:
i second jeffrey's suggestion, but otherwise the code and tests look to be wired up correctly.

cherewaty (approved):

PR #78, opened 2018-02-22
Implement Provider Sort Options

carolegodfrey:
## Purpose
Add support for `?sort=relevance` and `?sort=name` to `/eholdings/providers` endpoint
https://issues.folio.org/browse/UIEH-84

## Approach
- Support provider sort options which are translated into RM API `orderby` parameters. (relevance, vendorname) .
- Updates required to provider.rb and added associated tests with recordings.
- If a sort option is not supplied, it defaults to relevance (if a search is present) or vendorname (if a search is not present)


cherewaty (approved):
LGTM

PR #80, opened 2018-03-07
Update error messages to sentence case instead of title case

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-168, since FOLIO's standard is to use sentence case, modify error messages to use sentence case instead of title case. 

## Approach
- Since we cleaned up all error messages to be in sentence case, ui-eholdings can display them without further modification in ui.

## Screenshots
sentence_case_backend

### DIFF
--- a/app/models/configuration.rb
+++ b/app/models/configuration.rb
@@ -38,7 +38,7 @@ class Configuration
 
     return true if response.ok?
 
-    errors[:base] << 'RM-API Credentials Are Invalid'
+    errors[:base] << 'RM-API credentials are invalid'
   end
 
   def save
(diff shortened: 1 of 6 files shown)


carolegodfrey:
Looks good to me - just one question - should sentence casing apply to things like publication types for example  (Streaming audio) and content types (Aggregated full text) or not applicable?

cherewaty (approved):
@carolegodfrey yes, the publication types should probably be sentence case too. That can be another PR, since this one's good to go.

carolegodfrey (approved):

sduvvuri-ebsco:
Sure, thanks for your feedback Carole and Jeffrey! Will create another PR making that change!

PR #84, opened 2018-03-21
FOLIO-1155 Upgrade loofah dependency

cherewaty:
We received a notification from GitHub about vulnerabilities in `loofah < 2.2.1`.

Resolves https://issues.folio.org/browse/FOLIO-1155.


Robdel12 (approved):
Woop!

PR #85, opened 2018-03-22
UIEH-79 Allow editing of coverage statements

cherewaty:
## Purpose
On a customer resource, `coverageStatement` should be an editable field. It's simply a string.

Part of https://issues.folio.org/browse/UIEH-79

## Approach
- `coverageStatement` can only be not null if the customer resource `isSelected`.

### DIFF
--- a/spec/requests/customer_resources_spec.rb
+++ b/spec/requests/customer_resources_spec.rb
@@ -322,6 +322,36 @@ RSpec.describe 'Customer Resources', type: :request do
         end
       end
 
+      describe 'setting a coverage statement' do
+        let(:params) do
+          {
+            'data' => {
+              'type' => 'customerResources',
+              'attributes' => {
+                'isSelected' => false,
+                'visibilityData' => nil,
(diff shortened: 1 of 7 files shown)


spec/requests/customer_resources_spec.rb
```diff
@@ -467,6 +497,38 @@
...
+            put '/eholdings/customer-resources/22-1887786-1440285',
+                params: params, as: :json, headers: update_headers
+          end
+        end
+
+        it 'responds with OK status' do
+          expect(response).to have_http_status(200)
+        end
+
+        let!(:json) { Map JSON.parse response.body }
```
sadatay, line 524: this isn't directly related to this PR but i think we're using `let` way incorrectly throughout the suite.  i'm gonna make a note to go clean that up...

sadatay:
note: this may require ui-eholdings changes to the payload that gets sent when updating customer resources.

sadatay (approved):

elrickvm (approved):

PR #86, opened 2018-03-23
Implement Custom Labels backend to be displayed in Settings page in ui-eholdings; GET, PUT and DELETE

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-203, users should be able to view and edit custom labels from the Settings page in ui-eholdings. This work in the backend allows ui to get and put custom label related data. 

## Approach
- Added code to get custom label information from RM API. This also gets "root proxy" information and we prune it at Serializable level
- To get a list of custom labels from mod-kb-ebsco: GET /eholdings/custom-labels/
- Added code to put custom label data to RM API. RM API needs "root proxy" and all 5 custom labels data in a PUT call, if we do not submit all, then it resets them. From ui-eholdings, in order to be able to update on a per custom label basis, in mod-kb-ebsco, we do a GET call prior to PUT, prune the information from the GET to not contain any custom labels that have a null "displayLabel", patch the data from the request body of the PUT call to the information from GET and send it to RM API as a PUT. 
- To PUT custom label in mod-kb-ebsco: PUT /eholdings/custom-labels/{id}
- Sample request body for put request ```{
  "data": {
    "id": "1",
    "type": "customLabel",
    "attributes": {
      "id": 1,
      "displayLabel": "Hello first label testing",
      "displayOnFullTextFinder": true,
      "displayOnPublicationFinder": false
    }
  }
}```
- Added associated unit tests for the GET and PUT calls and took care of Rubocop suggestions.
- Added code for incoming param validation and its associated tests
- Added support for deleting a custom label, validations and its associated unit tests
- To delete a custom label using mod-kb-ebsco: DELETE /eholdings/custom-labels/{id}; no request body is required; if the deletion is successful, 204 is returned, if the label we are trying to delete does not exist, 404 is returned.

## Learning
- Flexirest API, specially, ResultIterator
- Ruby on Rails tutorials to understand routing
- Ruby syntax as and when needed

## Screenshots
custom_label_delete

### DIFF
--- a/app/controllers/custom_labels_controller.rb
+++ b/app/controllers/custom_labels_controller.rb
@@ -1,0 +1,57 @@
+# frozen_string_literal: true
+
+class CustomLabelsController < ApplicationController
+  def index
+    render jsonapi: custom_labels.all, include: params[:include]
+  end
+
+  def update # rubocop:disable Metrics/AbcSize
+    data_attributes = JSON.parse(request.body.read)['data']['attributes'] || {}
+    label_id = params[:id].to_i
+
+    custom_label_validation =
+      Validation::CustomLabelUpdateParameters.new(data_attributes, label_id)
+
+    if custom_label_validation.valid?
+      @custom_label = custom_labels.update(
+        label_id,
+        data_attributes
+      )
+      render jsonapi: @custom_label
+    else
+      render jsonapi_errors: custom_label_validation.errors,
+             status: :unprocessable_entity
+    end
+
+  # NoMethodError is raised when [] is invoked on 'data' or
+  # `data_attributes` and they are `nil`
+  rescue JSON::ParserError, NoMethodError
+    error = {
+      title: 'Invalid JSON',
+      detail: 'The provided JSON payload could not be parsed'
+    }
+
+    render jsonapi_errors: [error],
+           status: :unprocessable_entity
+  end
+
+  def destroy
+    label_id = params[:id].to_i
+
+    custom_label_validation =
+      Validation::CustomLabelDestroyParameters.new(label_id, custom_labels.all)
+
+    if custom_label_validation.valid?
+      custom_labels.delete(label_id)
+    else
+      render jsonapi_errors: custom_label_validation.errors,
+             status: :not_found
+    end
+  end
+
+  private
+
+  def custom_labels
+    CustomLabel.configure config
+  end
+end
--- a/app/controllers/validation/custom_label_destroy_parameters.rb
+++ b/app/controllers/validation/custom_label_destroy_parameters.rb
@@ -21,7 +21,7 @@ module Validation
       custom_label_list.each do |label|
         next unless label.id == label_id
         if label.display_label == ''
-          errors.add(:label_id, ':Label with this id does not exist')
+          errors.add(:label_id, ':Label with this id has already been deleted')
         end
       end
     end
--- a/app/controllers/validation/custom_label_destroy_parameters.rb
+++ b/app/controllers/validation/custom_label_destroy_parameters.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+module Validation
+  class CustomLabelDestroyParameters
+    include ActiveModel::Validations
+
+    attr_accessor :label_id, :custom_label_list
+
+    validates :label_id, presence: true
+    validate :label_id_range?
+    validate :label_deleted?
+
+    def label_id_range?
+      # there can be only 5 custom labels
+      errors.add(:label_id, ':Invalid custom label id') unless
+        label_id.to_i.between?(1, 5)
+    end
+
+    def label_deleted?
+      # check if the label has already been deleted
+      custom_label_list.each do |label|
+        next unless label.id == label_id
+        if label.display_label == ''
+          errors.add(:label_id, ':Label with this id does not exist')
+        end
+      end
+    end
+
+    def initialize(label_id, custom_label_list)
+      @label_id = label_id
+      @custom_label_list = custom_label_list
+    end
+  end
+end
--- a/app/controllers/validation/custom_label_update_parameters.rb
+++ b/app/controllers/validation/custom_label_update_parameters.rb
@@ -1,0 +1,52 @@
+# frozen_string_literal: true
+
+# rubocop:disable Naming/VariableName
+
+module Validation
+  class CustomLabelUpdateParameters
+    include ActiveModel::Validations
+
+    attr_accessor :label_id,
+                  :id,
+                  :displayLabel,
+                  :displayOnFullTextFinder,
+                  :displayOnPublicationFinder
+
+    validates :displayLabel, :id, presence: true
+    validate :ids_match?
+    validate :id_range?
+    validates :displayOnFullTextFinder,
+              inclusion: { in: [true, false], message: 'Invalid value' }
+    validates :displayOnPublicationFinder,
+              inclusion: { in: [true, false], message: 'Invalid value' }
+
+    def ids_match?
+      # label_id is what is passed in the url
+      # id is what is passed in the payload
+      # this should be done in Rails 5 using the following syntax
+      # validates :id, numericality: { only_integer:true }
+      # But unfortunately, it does not work, hence this is needed
+      errors.add(:id, ':Label ids should match') unless
+        (id.is_a? Integer) && (id == label_id)
+    end
+
+    def id_range?
+      # there can be only 5 custom labels
+      errors.add(:id, ':Invalid custom label id') unless
+        id.to_i.between?(1, 5)
+    end
+
+    def initialize(params, label_id)
+      @label_id = label_id
+      # For consistency, this should be an integer
+      @id = params['id']
+      @displayLabel = params['displayLabel']
+      @displayOnFullTextFinder =
+        params['displayOnFullTextFinder']
+      @displayOnPublicationFinder =
+        params['displayOnPublicationFinder']
+    end
+  end
+end
+
+# rubocop:enable Naming/VariableName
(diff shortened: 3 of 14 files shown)


app/models/custom_label.rb
```diff
@@ -0,0 +1,67 @@
...
+    @attrs['displayOnPublicationFinder']
+  end
+
+  def self.all
+    rmapi_customer_root.all.labels.map { |attrs| new attrs }
+  end
+
+  def self.update(label_id, attrs)
+    id = label_id.to_i
+    return unless id.positive?
```
cowboyd: We shouldn't have different return values from a single method. In other words, we should still return a custom label no matter what, not `nil` in one case, and `new attrs` in another.

app/models/custom_label.rb
```diff
@@ -0,0 +1,67 @@
...
+    root = rmapi_customer_root.all
+    # prune the root to not include labels with displayLabel=""
+    # because RM API would throw an error in PUT request
+    prune_root(root, id)
```
cowboyd: I'd move all of the pruning / patching / updating into the `RmapiCustomerRoot` since that is the class that ultimatley has knowledge of the RMAPI and abstracts away how it works. In other words, if the RMAPI changes its contract, then only this class should need to change.

So here we would say:
```ruby
root.update_label(label_id, attrs)
return new attrs
```
and then in `RmApiCustomerRoot`

```ruby
def update_label(label_id, attrs)
  prune(label_id)
  patch(label_id, attrs)
  update()
end
```

app/models/custom_label.rb
sduvvuri-ebsco: This was more like a defense mechanism, in case the consumer provides some invalid value for the update like -1. was thinking more on lines of logging and reporting an error that the provided value is invalid. Are you suggesting we just return an empty response without any custom labels instead?

app/models/custom_label.rb
sduvvuri-ebsco: Sure, thanks for your suggestion! I will move it..

app/models/custom_label.rb
sduvvuri-ebsco: Ok, realized that validations could be done in the controller.. will update code accordingly

sduvvuri-ebsco:
This job is failing in Jenkins of IndexData and has something to do with their RVM script location. @funkymalc is looking at it.

app/controllers/custom_labels_controller.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+class CustomLabelsController < ApplicationController
+  def index
+    render jsonapi: custom_labels.all, include: params[:include]
+  end
+
+  def update
+    data_attributes = JSON.parse(request.body.read)['data']['attributes'] || {}
```
sadatay: you're gonna want to throw a `rescue` at the bottom of the update method to catch any `JSON:ParserError` that may be thrown due to a malformed request body.  there's an example of this in `configurations_controller.rb` that should be pretty easy to port over

app/controllers/custom_labels_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+    validate(data_attributes, label_id)
+  end
+
+  private
+
+  def custom_labels
+    CustomLabel.configure config
+  end
+
+  def validate(data_attributes, label_id)
```
sadatay: typically we try to follow a format that is roughly:

- do validation
  - if valid
    - do update and render result
  - else
    - render an error

example from packages_controller:
```ruby
  def update
    package_validation = Validation::PackageParameters.new(package_params)

    if package_validation.valid?
      @package.update package_params
      render jsonapi: @package
    else
      render jsonapi_errors: package_validation.errors,
             status: :unprocessable_entity
    end
  end
```

so you can probably just hoist the contents of `validate` back up inline with the `update` method.  only possible obstacle might be rubocop carping about cyclomatic complexity, but we can work around that.  my main pain point here is that `validate` does more than just validation

app/models/custom_label.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  def display_on_publication_finder
+    @attrs['displayOnPublicationFinder']
+  end
+
+  def self.all
+    rmapi_customer_root.all.labels.map { |attrs| new attrs }
+  end
+
+  def self.update(id, attrs)
```
sadatay, line 28: ~~this model represents a single instance of a custom label, right?  in that case i don't think you need to pass the id here, since the model should already have an id (via the method on line 8).~~  

nvm, i see what's up.  clever 😄

app/controllers/custom_labels_controller.rb
sduvvuri-ebsco: Yep, rubocop was complaining about complexity of the update method which is why I had to split it up the way I did. How do we get around rubocop complaining, do you mean by disabling its concern about complexity using `rubocop:disable Metrics/AbcSize`

app/controllers/custom_labels_controller.rb
sduvvuri-ebsco: sure, yes, good point, thanks!

app/models/custom_label.rb
```diff
@@ -0,0 +1,46 @@
...
+    root = rmapi_customer_root.all
+    root.update_label(id, attrs)
+    # return updated custom label
+    new attrs
+  end
+
+  def self.configure(config)
```
sadatay: ✨ gettin' meta up in here ✨

app/models/rm_api_customer_root.rb
```diff
@@ -0,0 +1,31 @@
...
+    # update the root
+    update
+  end
+
+  def prune(id)
+    labels.delete_if { |item| item.id != id && item.displayLabel == '' }
+  end
+
+  def patch(id, attrs)
+    labels.each do |label|
```
sadatay, line 24: this is acting on a singleton label right?  instead of the `each` you could probably do something like `label = labels.find { |label| label.id == id }` and then set your data attributes on that.  not sure if it's any more efficient (maybe even less efficient?) but i think it's easier to read.

it'd be nice if there were some way to map the contents of `attrs` right onto the label object too... i bet there is but i can't think of it off the top of my head.  i'll look into it and get back to you

sduvvuri-ebsco: Sure thing. 
Regarding the usage of "each" vs. "find", I think each is more efficient for traversing an array and replacing content in place, "find" returns a new array with the replaced values.

carolegodfrey (approved):
Looks Great! Really helpful documentation and tests!

app/controllers/validation/custom_label_update_parameters.rb
```diff
@@ -0,0 +1,52 @@
...
+              inclusion: { in: [true, false], message: 'Invalid value' }
+    validates :displayOnPublicationFinder,
+              inclusion: { in: [true, false], message: 'Invalid value' }
+
+    def ids_match?
+      # label_id is what is passed in the url
+      # id is what is passed in the payload
+      # this should be done in Rails 5 using the following syntax
+      # validates :id, numericality: { only_integer:true }
+      # But unfortunately, it does not work, hence this is needed
```
sadatay, line 28: >But unfortunately, it does not work

any info on why that is?  this does seem a bit strange, but jsonapi-rb has definitely pulled weirder moves before about stuff like this.  i'd be curious to read up on why/how this is a rails 5 thing and why our rails 5 app isn't in line with that.  not blocking the ticket or anything, just curious if you had any resources on this before i google around :)

app/controllers/validation/custom_label_destroy_parameters.rb
```diff
@@ -0,0 +1,34 @@
...
+      errors.add(:label_id, ':Invalid custom label id') unless
+        label_id.to_i.between?(1, 5)
+    end
+
+    def label_deleted?
+      # check if the label has already been deleted
+      custom_label_list.each do |label|
+        next unless label.id == label_id
+        if label.display_label == ''
+          errors.add(:label_id, ':Label with this id does not exist')
```
sadatay: so if i understood you earlier when we spoke, it sounds like we are in fact showing all custom labels to api consumers on GET to this endpoint.  i agree we want to catch this case and respond to it, but it feels weird saying it was "deleted" here, if a label with that id _does_ technically "exist" in their view were they to run a GET against this endpoint.  is that true or did i miss something?  

if that is the case how would you feel about changing this to "label has already been destroyed" or "label is disabled" - something less explicit basically?

app/models/rm_api_customer_root.rb
sadatay: k,that makes sense 👍

app/controllers/validation/custom_label_update_parameters.rb
sduvvuri-ebsco: The way I figured that it is not working as I intended is by testing it manually in my code. ```validates :id, numericality: { only_integer:true }``` was what I tried and even if I pass a string as id, for example "abc",  it was not throwing an error. I did google about it and read somewhere that it is a regression in Rails 5, but am not able to find that link now :(

sadatay:
lookin' real good 👍 and i agree with carole on the tests and the super comprehensive PR description.  

we should throw down on some general mod-kb and ui-eholdings docs one of these iterations - at least for the dev experience of anybody coming on board in the future

sadatay (approved):
:shipit:

app/controllers/validation/custom_label_destroy_parameters.rb
sduvvuri-ebsco: mm, yea, I know ....so per RM API, a custom label is considered non-existing if the "displayLabel" attribute is "", sure, our consumers would not know that as it is not intuitive.. I can change the error message to say that "label with this id has been deleted already".

PR #87, opened 2018-04-02
UIEH-218 Search titles within a detail record

elrickvm:
## Purpose
 - Resource Librarians after searching for a Package want to be able to then search titles that are within that Package.

## Approach
 - Pass along all necessary parameters to allow the appropriate query to be constructed.
 - Map query search and filter options sent from UI to what RM-API expects to receive to this endpoint

## Jira
https://issues.folio.org/browse/UIEH-218


cherewaty:
LGTM 👍

app/models/customer_resource.rb
```diff
@@ -20,12 +20,41 @@ class CustomerResource < RmApiResource
...
+      request.get_params[:orderby] =
+        if sort == 'relevance'
+          'relevance'
+        elsif sort == 'name'
+          'titlename'
+        else
+          request.get_params[:q] ? 'relevance' : 'titlename'
+        end
+
+      request.get_params[:search] = request.get_params[:q] || ''
```
wwilsman: `get_params.delete(:q)` will return the `:q` param, so instead of deleting it below, you could delete it here.

``` ruby
request.get_params[:search] = request.get_params.delete(:q) || ''
```

elrickvm: @sadatay had the same sentiment. I tried that @wwilsman but it was failing the test. Will give it another shot now and see if was because of something else i was doing.

app/models/customer_resource.rb
```diff
@@ -20,9 +20,37 @@ class CustomerResource < RmApiResource
 
   before_request do |name, request|
     if name == :find_by_package
-      request.get_params[:search] = ''
+      filters = request.get_params.delete(:filter) || {}
+
+      unless filters.is_a?(ActionController::Parameters) || filters.is_a?(Hash)
```
cowboyd, line 25: Is this necessary? As far as I can tell, this is cranked through the `**` operator here, so I don't see how this could ever be anything but a `Hash`

app/models/customer_resource.rb
```diff
@@ -20,9 +20,37 @@ class CustomerResource < RmApiResource
...
+      request.get_params[:orderby] =
+        if sort == 'relevance'
+          'relevance'
+        elsif sort == 'name'
+          'titlename'
+        else
+          request.get_params[:q] ? 'relevance' : 'titlename'
+        end
+
+      request.get_params[:search] = request.get_params.delete(:q) || ''
```
sadatay, line 52: i _knew_ this worked, it must've been failing for other reasons.  this has been an idiom since at least perl. 

👀 except javascript apparently.  javascript would return `true`.

cowboyd (commented):
In general the biggest concern I have with this is that it doesn't really capture what all of the munging of parameter is doing. 

If there is any way to avoid editing in-plae of the `request` i.e. mutating it, I would. 

It's very unclear what this anonymous `before_request` hook is doing, so I'd have some more documentation about each operation.

It appears that it is munging parameters to make them suitable for presentation to RMAPI. In that case, it might help to say:

```ruby
request.get_params.merge! params_for_rmapi(request)
request.get_params.delete_if(&non_rmapi_param)
```

app/models/customer_resource.rb
wwilsman: `filter` is a named param. You can pass `?q=search&filter=blah` and it would work up until we try to pull a value off of `filter`. This was first introduced when `filter` briefly did accept a string instead of a hash.

`find_by_package(**params)` is the same as `find_by_package(q: ..., filter: ...)`. So the `filters.is_a?` check ensures `filter` is a nested param

app/models/customer_resource.rb
```diff
@@ -20,9 +20,37 @@ class CustomerResource < RmApiResource
 
   before_request do |name, request|
     if name == :find_by_package
-      request.get_params[:search] = ''
+      filters = request.get_params.delete(:filter) || {}
```
sadatay, line 23: so this is probably my bad for setting the precedent many 🌔 ago when i was first learning my way around Flexirest: but we really, really need to give these poor `before_request` blocks a makeover. stack much more in here and it's liable to collapse, destroying nearby files in the process.  we could at the very least pull each section into a named method, but looking at how much repetition there is here i'm sure it could go way beyond that.  this, or one of its cousins, is in 7 files @ around 50 lines each.

since that's not really in the scope of the ticket, i'm not gonna hold you accountable for a bad trend i started that has seemingly taken off in every model in the codebase.  but i think it's worth noting it, lest we forget to someday address this.

app/models/customer_resource.rb
elrickvm: You are correct in that it is 'munging' params. The `params_for_rmapi` may work. But does it matter that what params are passed to RMAPI is dependent upon wether its a Package, Package-Title?

cowboyd (approved):
Given the fights that we've been having trying to decorate the parameters using FlexiRest, it's becoming clear that we're outgrowing it and that it needs to be replaced.

We're going to merge as-is, and then we'll open up a epic / story for moving away from flexures altogether.

PR #88, opened 2018-04-02
Implement GET and PUT for root proxy to be used from Settings in ui-eholdings

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-219, users should be able to get the root proxy list to be displayed in drop-down in Settings of ui-eholdings, and they should be able to update the root proxy that is currently set. 

## Approach
- Implement the GET call /eholdings/root-proxies -- This gives a list of root proxies with params from RM API for each proxy like "id", "name" and "urlMask" as well as "selected", which allows ui-eholdings to highlight selection in drop down.
- This is a aggregated result of 2 calls to RM API
- By calling GET /{custid}/proxies in RM API -- gives the list of proxies available for a customer
- By calling GET /{custid} -- gives the current root proxy and custom labels
- Aggregate both these results, filter results and provide a response for GET
- Implement a PUT call /eholdings/root-proxies/id -- This allows user to update the current selection of root proxy
- Request body schema: ```{
  "data": {
    "id": "<n>",
    "type": "rootProxy",
    "attributes": {
      "id": "<n>"
    }
  }
}```
- Validations are done such that id in url should match id in request body; and id passed in should be one of the allowed root proxies for the customer; else error messages are returned
- If it passes validations, then update is made to the root proxy; the updated value with "selected" attribute set to true is returned in response. "name" and "urlMask" are set to null since they are not updated. If we are to populate "name" and "urlMask" with correct values, we need to make another call to RM API to get their real values and merge it in with the result. We are not making that additional GET call to save traffic to RM API for data that we might not use. They are in the payload for consistency between GET and PUT requests.
- In order to update in RM API, we get GET /{custid}, prune the response to ensure that custom labels with "displayLabel" equal to empty do not exist, patch the response with the updated proxy and make a PUT call to /{custid} with combined payload of custom labels and root proxy.
- Added associated unit tests. 

## Todo
- Confirm with Khalilah if "urlMask" is needed; remove it from GET and PUT requests if not needed. 

## Screenshots
root_proxies

### DIFF
--- a/ModuleDescriptor.json
+++ b/ModuleDescriptor.json
@@ -19,6 +19,10 @@
           "pathPattern": "/eholdings/custom-labels"
         },
         {
+          "methods": [ "GET", "PUT" ],
+          "pathPattern": "/eholdings/root-proxies"
+        },
+        {
           "methods": [ "GET" ],
           "pathPattern": "/eholdings/vendors*"
         },
--- a/app/controllers/root_proxies_controller.rb
+++ b/app/controllers/root_proxies_controller.rb
@@ -1,0 +1,43 @@
+# frozen_string_literal: true
+
+class RootProxiesController < ApplicationController
+  def index
+    render jsonapi: root_proxies.all, include: params[:include]
+  end
+
+  def update # rubocop:disable Metrics/AbcSize
+    data_attributes = JSON.parse(request.body.read)['data']['attributes'] || {}
+    root_proxy_id = params[:id]
+
+    root_proxy_validation =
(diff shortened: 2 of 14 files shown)


cherewaty (approved):
All makes sense to me 👍 Since this was a big, tricky one, I recommend >=1 more review.

carolegodfrey (approved):
LGTM - I agree with not making the added RM API call to populate name and urlMask for the PUT Response - nicely documented!

PR #89, opened 2018-04-03
UIEH-230 Add Display and Edit Provider Level Tokens

carolegodfrey:
# Purpose
[UIEH-230](https://issues.folio.org/browse/UIEH-230)
Some Providers require a token to be able to access content. This PR adds support for 
- Getting a Provider's Token Details
- Updating a Provider's Token Value

## Approach
- Update Provider GET method to include additional details from RM API response. Specifically a providerToken field with attributes such as the following is returned in the GET response attributes collection:
`"providerToken": {
        "factName": "[[testfact]]",
        "prompt": "/test/",
        "helpText": "test test",
        "value": "12345"
    }`
- Providers that do not support tokens return a `nil` providerToken field
- Added support for a Provider PUT method to allow update of providerToken `value` field. Followed similar pattern as is available for package and customer_resource PUT requests.
- Sample Provider PUT has the following body format ` {"data": {"type": "providers","attributes": {"providerToken": {"value":"999999"}}}}`
- RM API allows for a maximum length token of 500 characters (larger fields are accepted but truncated). 
- Added token field validation for a maximum of 500 characters
- There are no restrictions on the characters that can be entered in the token field
- Added mappings in serializable and deserializable to handle vendor to provider mappings
- Added associated unit tests

#### TODOS and Open Questions
- [ ] Error responses from RM API `Vendor` requests currently contain lower case keys in body, such as below
`{"errors":[{"code":1001,"subcode":0,"message":"Vendor not found"}]}`. 
which differs from currently expected format 
`{"Errors":[{"Code":1001,"Subcode":0,"Message":"Vendor not found"}]}`.
As a result, error message text (as a result of an RM API returned error) is not currently returned in error response body. 
To address as separate PR

## Learning
Ruby, Rails, JSON-api
http://jsonapi-rb.org/guides/getting_started/
http://api.rubyonrails.org/classes/ActionController/Parameters.html
http://edgeguides.rubyonrails.org/api_app.html

### DIFF
--- a/ModuleDescriptor.json
+++ b/ModuleDescriptor.json
@@ -19,7 +19,7 @@
           "pathPattern": "/eholdings/vendors*"
         },
         {
-          "methods": [ "GET" ],
+          "methods": [ "GET", "PUT" ],
           "pathPattern": "/eholdings/providers*"
         },
         {
--- a/app/controllers/providers_controller.rb
+++ b/app/controllers/providers_controller.rb
@@ -54,7 +54,7 @@ class ProvidersController < ApplicationController
     params
       .require(:provider)
       .permit(
-        providerToken: [:value]
+        vendorToken: [:value]
       )
   end
 end
--- a/app/controllers/providers_controller.rb
+++ b/app/controllers/providers_controller.rb
@@ -1,7 +1,10 @@
 # frozen_string_literal: true
 
 class ProvidersController < ApplicationController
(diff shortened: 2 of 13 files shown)


sduvvuri-ebsco (approved):
Looks great Carole, and as we discussed we should create a user story for cleaning up vendor route in mod-kb-ebsco that was supported a while ago!

PR #91, opened 2018-04-04
UIEH-230 - Add support for both Errors and errors from RM API response

carolegodfrey:
## Purpose
[UIEH-230](https://issues.folio.org/browse/UIEH-230)
The RM API `vendors` endpoint is returning an error message body which is inconsistent. Specifically a different case of keys is returned in the  JSON error response body. 
For example: 
`{"errors":[{"code":1001,"subCode":0,"message":"Vendor not found"}]}`
When this is expected:
`{"Errors":[{"Code":1001,"Message":"Vendor not found","SubCode":0}]}`
Because of this, we do not see  RM API error message text in mod-kb-ebsco error response body.

## Approach
- Update error handling in mod-kb-ebsco so that RM API message text in either format is available in response body from mod-kb-ebsco. 
- Made changes in application controller to handle both cases of error response. 
- Updated provider unit tests to verify RM API error message text
- Replace vendor with provider in error response text

#### TODOS and Open Questions
- [ ]  This is a temporary solution as longer term, casing will be resolved and consistent in RM API
- [x]  ~~RM API error text includes terminology of vendor vs provider - should this be replaced in text itself~~
- [ ]  Is there a better location in code for a mapping such as this

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -39,18 +39,30 @@ class ApplicationController < ActionController::API
            status: :bad_request
   end
 
-  def catch_flexirest_exceptions
+  def catch_flexirest_exceptions # rubocop:disable Metrics/AbcSize
     yield
   rescue Flexirest::HTTPClientException,
          Flexirest::HTTPServerException,
          Flexirest::HTTPNotFoundClientException => e
 
-    errors_hash = e.result.Errors.to_a.map do |err|
-      { "title": err.to_hash['Message'] }
-    end
+    errors_hash = if e.result.respond_to?(:Errors)
+                    e.result.Errors.to_a.map do |err|
+                      { "title": map_provider(err.to_hash['Message']) }
+                    end
+                  elsif e.result.respond_to?(:errors)
+                    e.result[:errors].items.to_a.map do |err|
+                      { "title": map_provider(err.to_hash['message']) }
+                    end
+                  else
+                    []
+                  end
 
     render jsonapi_errors: errors_hash,
            status: e.status
+  end
+
+  def map_provider(string)
+    string.gsub(/Vendor/, 'Provider').gsub(/vendor/, 'provider')
   end
 
   def verify_okapi_headers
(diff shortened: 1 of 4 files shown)


app/controllers/application_controller.rb
```diff
@@ -39,20 +39,32 @@ def catch_exceptions
...
+  def catch_flexirest_exceptions # rubocop:disable Metrics/AbcSize
     yield
   rescue Flexirest::HTTPClientException,
          Flexirest::HTTPServerException,
          Flexirest::HTTPNotFoundClientException => e
 
-    errors_hash = e.result.Errors.to_a.map do |err|
-      { "title": err.to_hash['Message'] }
-    end
+    errors_hash = if e.result.respond_to?(:Errors)
```
cowboyd: I'd go ahead and just extract this into a method. It'll fix the formatting, and make it clearer, and (I think) remove the need to disable the lint warning.

spec/requests/providers_spec.rb
```diff
@@ -342,8 +345,13 @@
           end
         end
 
+        let!(:json) { Map JSON.parse response.body }
+
         it 'responds with bad request' do
+          # rubocop:disable Metrics/LineLength
```
cowboyd: If this rubocop rule is too onerous, I'd rather take it out than one-off disable it.

cowboyd:
Also, are there any backwards compatibility concerns here? As in, do we need to have both versions of the error messages hash to bridge upgrading the client?

app/controllers/application_controller.rb
carolegodfrey: Thanks for the suggestion - moved out to a method (is definitely cleaner) - but still had the lint warning

spec/requests/providers_spec.rb
carolegodfrey: I disabled this check and removed other references (primarily in tests)

carolegodfrey:
On backwards compatibility - one production and sandbox rm api instances have been corrected we should be good to remove the temporary lower case version of errors, but definitely a good point to consider - thanks @cowboyd

cowboyd (approved):

PR #94, opened 2018-04-11
UIEH-262 Fix title list loading for custom packages

cherewaty:
## Purpose
The list of related customer resources for a custom package was not loading.

Resolves https://issues.folio.org/browse/UIEH-262.

## Approach
Custom titles don't return an `identifiersList` from the EBSCO RM API, so the serializer was getting hung up.


Robdel12 (approved):
This LGTM!

PR #95, opened 2018-04-11
UIEH-283 Return packageType from RM API

cherewaty:
## Purpose
We want to show a package's type (Complete, Partial, Variable, or Custom).

Needed for https://issues.folio.org/browse/UIEH-283

## Approach
Pass along `packageType` from RM API as-is.


Robdel12 (approved):
📦 👍

PR #96, opened 2018-04-12
UIEH-280 Enable editing for custom packages

cherewaty:
## Purpose
Enables and sets up test coverage for editing of a custom package's:
- name
- content type
- coverage dates
- visibility

Part of https://issues.folio.org/browse/UIEH-280

## Approach
`contentType` de/serialization is completely handled by `mod-kb-ebsco`. `mod-kb-ebsco` consumers send and receive human readable strings like "E-Book" and "Online Reference". EBSCO's RM API sends and receives "ebook" and "onlinereference".

This does not address a custom package's holding status or creating a custom package.

### DIFF
--- a/app/controllers/packages_controller.rb
+++ b/app/controllers/packages_controller.rb
@@ -65,6 +65,8 @@ class PackagesController < ApplicationController
     params
       .require(:package)
       .permit(
+        :packageName,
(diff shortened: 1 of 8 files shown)


Robdel12 (approved):
LGTM!

PR #98, opened 2018-04-13
UIEH-293 Enable editing for custom titles

cherewaty:
## Purpose
Enables some fields to be editable on custom titles ("custom titles" are really resources, a relationship between a package and a title).

https://issues.folio.org/browse/UIEH-293

## Approach
The biggest challenge was working around the `customerResourcesList`. We hadn't previously been updating fields with `PUT /eholdings/customer-resources/:id` that weren't inside that nested list. `titleName` and `pubType` are at the top level.

Now that we're creating essentially "custom customer resources", we should consider shortening "customer resources" in our vocabulary to simply "resources".

The fields that are tested and editable with this PR:
- Visibility
- Title name
- Publication type
- Coverage dates
- Coverage statement
- Embargo period

_Not_ in this PR:
- Peer review
- Publisher
- Edition
- Contributors
- Identifiers
- Description
- Proxy
- Custom URL
- Coverage selection (custom coverage or coverage statement)
- Custom labels
- Adding a title to a custom package

### DIFF
--- a/app/controllers/customer_resources_controller.rb
+++ b/app/controllers/customer_resources_controller.rb
@@ -48,6 +48,8 @@ class CustomerResourcesController < ApplicationController
     params
       .require(:customer_resource)
       .permit(
+        :titleName,
+        :pubType,
         :isSelected,
         :coverageStatement,
         visibilityData: [:isHidden],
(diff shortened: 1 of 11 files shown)


Robdel12 (approved):
Looks good to me!

carolegodfrey (approved):

PR #99, opened 2018-04-13
Does a provider support custom packages?

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-291, RM API vendor provides a flag which lets us know whether a vendor allows custom packages to be added.  

## Approach
Map the field "isCustomer" holding a boolean value provided by RM API to "supportsCustomPackages" in mod-kb-ebsco and expose it from GET calls for provider so ui-eholdings can pick it up.

## Screenshots
supprtscustompackages


carolegodfrey (approved):
Looks good!

PR #100, opened 2018-04-13
Rename customer resources to just resources

cherewaty:
## Purpose
Now that we're beginning to deal with custom titles (which are really a custom "package-title", we'd end up with "custom customer resources".

"Customer resources" was a term used by RM API, but is not a common way to describe that relationship.

## Approach
To eliminate developer confusion, settle on simply "resources" instead of "customer resources".

Changes endpoints and references in the code.


sduvvuri-ebsco (approved):
LGTM!

carolegodfrey (commented):
Looks good - will help with name tongue twister - any concerns though with backward compatibility? Updates for this need to happen to both mod-kb-ebsco and ui-eholdings in unison

cherewaty:
@carolegodfrey no concerns with compatibility, we're ready on the other end too :) https://github.com/folio-org/ui-eholdings/pull/347

cowboyd:
Are we going to accept some downtime with this one?

cherewaty:
@cowboyd since we have the PR ready for both repos, I think the 10-20 min of downtime will be fine.

cowboyd (approved):

app/controllers/titles_controller.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class TitlesController < ApplicationController
-  before_action :set_title, only: %i[show customer_resources]
+  before_action :set_title, only: %i[show resources]
```
cowboyd, line 4: As a general note (and completely out of scope of this PR), `before_action`s are overkill here. A simple method is just fine:

```ruby
def title
  titles.find params[:id]
end
```

cowboyd: And since only the `show` and `resources` actions will every call it, the `only: ` clause enforced simply via usage.

PR #101, opened 2018-04-13
[Part of UIEH-280] POST to create a custom package

Robdel12:
## Purpose
Per https://issues.folio.org/browse/UIEH-280, users need the ability to create a custom package through ui-eholdings. This implementation solves that issue. 

## Approach
- Expect UI to send "name"(packageName) and "contentType" attributes which are mandatory for RM API to be able to create a custom package. UI can also send "customCoverage" which includes "beginCoverage" and "endCoverage".
- Validate the presence of "name" and "contentType"
- RM API supports only a set of "contentTypes" -- validate that content type passed in is part of them, else map it to "Unknown"
- If all validations mentioned above pass, request RM API for "vendorId" corresponding to the "customer id" for a given tenant making the request.
- Use the "vendorId" as well as other attributes mentioned above to make a POST to RM API
- If POST is successful, RM API gives the "packageId" of newly created package.
- Make another GET call to RM API to get the details of newly created package, merge it with the "packageId" and give that as response for the POST request to mod-kb-ebsco
- Validations can fail for below reasons: 
   - "name" attribute missing - response 422
   - "contentType" missing - response 422
   - "package name already exists" - response 400
- Sample POST request - POST /eholdings/packages/
- Sample request body -
`{
  "data": {
    "type": "packages",
    "attributes": {
      "packageName": "SD test new custom package",
      "contentType": "Print"
    }
  }
}`
- Expected response code - 200
- Associated unit tests
 
## Screenshots
post_custom_package

### DIFF
--- a/app/controllers/packages_controller.rb
+++ b/app/controllers/packages_controller.rb
@@ -1,9 +1,9 @@
 # frozen_string_literal: true
 
 class PackagesController < ApplicationController
-  before_action :set_package, only: %i[show update resources]
+  before_action :set_package, only: %i[show update destroy resources]
 
-  deserializable_resource :package, only: :update,
+  deserializable_resource :package, only: %i[create update],
                                     class: DeserializablePackage
 
   def index
@@ -16,6 +16,25 @@ class PackagesController < ApplicationController
 
     render jsonapi: @packages.packagesList.to_a,
            meta: { totalResults: @packages.totalResults }
+  end
+
+  def create
+    # Pass only those parameters that are allowed to be modified for
+    # custom packages
+    package_create_params = package_params.slice(
+      :packageName,
+      :contentType,
+      :customCoverage
+    )
+    package_validation = Validation::CustomPackageParameters.new(package_create_params)
+
+    if package_validation.valid?
+      @package = packages.create_package(package_create_params)
+      render jsonapi: @package
+    else
+      render jsonapi_errors: package_validation.errors,
+             status: :unprocessable_entity
+    end
   end
 
   def show
@@ -31,6 +50,17 @@ class PackagesController < ApplicationController
     else
       render jsonapi_errors: package_validation.errors,
              status: :unprocessable_entity
+    end
+  end
+
+  def destroy
+    package_validation = Validation::PackageDestroyParameters.new(@package)
+
+    if package_validation.valid?
+      @package.delete
+    else
+      render jsonapi_errors: package_validation.errors,
+             status: :bad_request
     end
   end
 
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -14,7 +14,7 @@ Rails.application.routes.draw do
       end
     end
 
-    resources :packages, only: %i[index show update] do
+    resources :packages, only: %i[create index show update destroy] do
       member do
         get 'resources'
       end
--- a/spec/requests/packages_spec.rb
+++ b/spec/requests/packages_spec.rb
@@ -1094,4 +1094,197 @@ RSpec.describe 'Packages', type: :request do
       end
     end
   end
+
+  describe 'deleting a custom package' do
+    describe 'delete a custom package successfully' do
+      before do
+        VCR.use_cassette('delete-custom-package') do
+          delete '/eholdings/packages/123355-2848230',
+                 headers: okapi_headers
+        end
+      end
+
+      it 'gets a successful response' do
+        expect(response).to have_http_status(204)
+      end
+    end
+
+    describe 'trying to delete a deleted package results in error' do
+      before do
+        VCR.use_cassette('delete-deleted-custom-package') do
+          delete '/eholdings/packages/123355-2848230',
+                 headers: okapi_headers
+        end
+      end
+
+      it 'gets a not found response' do
+        expect(response).to have_http_status(404)
+      end
+    end
+
+    describe 'trying to delete a non-custom package' do
+      before do
+        VCR.use_cassette('delete-non-custom-package') do
+          delete '/eholdings/packages/583-4345',
+                 headers: okapi_headers
+        end
+      end
+
+      it 'gets a bad request response' do
+        expect(response).to have_http_status(400)
+      end
+    end
+  end
+
+  describe 'creating a custom package' do
+    let(:update_headers) do
+      okapi_headers.merge(
+        'Content-Type': 'application/vnd.api+json'
+      )
+    end
+
+    describe 'without a packageName' do
+      let(:params) do
+        {
+          "data": {
+            "type": 'packages',
+            "attributes": {
+              "contentType": 'ebook'
+            }
+          }
+        }
+      end
+
+      before do
+        VCR.use_cassette('post-custom-package') do
+          post '/eholdings/packages/',
+               params: params, as: :json, headers: update_headers
+        end
+      end
+
+      it 'returns an error' do
+        expect(response).to have_http_status(422)
+      end
+    end
+
+    describe 'without a contentType' do
+      let(:params) do
+        {
+          "data": {
+            "type": 'packages',
+            "attributes": {
+              "name": 'totally new name'
+            }
+          }
+        }
+      end
+
+      before do
+        VCR.use_cassette('post-custom-package') do
+          post '/eholdings/packages/',
+               params: params, as: :json, headers: update_headers
+        end
+      end
+
+      it 'returns an error' do
+        expect(response).to have_http_status(422)
+      end
+    end
+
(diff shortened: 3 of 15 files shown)


app/controllers/packages_controller.rb
```diff
@@ -1,9 +1,11 @@
 # frozen_string_literal: true
 
+VENDOR_ID = '123355'
```
Robdel12: If we keep it hard coded I'd like to update this to be specific. Whose vendor ID is it?

cherewaty: We won't be able to have that hardcoded. Every library would have a different vendor ID.

Robdel12: Hm, this comment on the ticket leads me to think differently: 
> Our "custid" is "apidvcorp" and "vendorid" is "123355" --> The "custid" and "vendorid" do not change --> so we can hard-code them in mod-kb-ebsco. If we want to keep the API generic, we should actually be making a call to RM API to get "vendorid" as mentioned in additional notes section below.

cherewaty: I think that's since been edited in the issue.

sduvvuri-ebsco: Sorry, @Robdel12 , I updated the ticket after thinking it through to not have it hard-coded. I assumed you would get updates, so did not bother pinging you on Slack, my bad. There is additional notes in the ticket on how to retrieve the vendor id specific to a vendor.

app/controllers/packages_controller.rb
```diff
@@ -61,6 +82,24 @@ def packages
     Package.configure(config)
   end
 
+  # Smooth over slicing out only the info RM API needs to create a
+  # custom package This also passes along the required hardcoded
+  # VENDOR_ID
+  #
+  # Open questions:
+  # - Is it safe to hard code the VENDOR_ID like that?
+  # - What about the customer ID? Is that supposed to be hard coded?
```
sduvvuri-ebsco: customer id is in config already and is not hard-coded.

app/controllers/packages_controller.rb
```diff
@@ -1,9 +1,11 @@
 # frozen_string_literal: true
 
+VENDOR_ID = '123355'
+
 class PackagesController < ApplicationController
   before_action :set_package, only: %i[show update customer_resources]
 
-  deserializable_resource :package, only: :update,
+  deserializable_resource :package, only: %i[update create],
```
sduvvuri-ebsco: I am really being nitpicky here but can we please have the create before the update?

config/routes.rb
```diff
@@ -14,7 +14,7 @@
       end
     end
 
-    resources :packages, only: %i[index show update] do
+    resources :packages, only: %i[index show update create] do
```
sduvvuri-ebsco: a similar very nitpicky comment here: can we please have create before index for consistency?

sduvvuri-ebsco:
Oh, one other thing I just noticed, we should also add this new method in ModuleDescriptor.json in the corresponding path

app/controllers/packages_controller.rb
```diff
@@ -61,6 +81,27 @@ def packages
...
+      :contentType,
+      :customCoverage
+    )
+
+    { vendor_id: @provider.vendorId }.merge(relevent_rm_api_data)
+  end
+
+  def find_provider
+    provider = Provider.configure(config)
+                 .all({search: config.customer_id })
```
Robdel12: `{ search: config.customer_id }` (formatting)

app/controllers/packages_controller.rb
```diff
@@ -61,6 +81,27 @@ def packages
     Package.configure(config)
   end
 
+  # Smooth over slicing out only the info RM API needs to create a
+  # custom package This also passes along the required provider id
+  def rm_api_package
+    find_provider
+
+    relevent_rm_api_data = package_params.slice(
```
sadatay: relevant*.  also maybe just drop that part entirely from the variable name, imo.  it's relevant by virtue of the fact that we're doing something with it.

app/controllers/packages_controller.rb
```diff
@@ -18,6 +18,26 @@ def index
...
   end
 
+  def create
+    package_validation = Validation::CustomPackageParameters.new(package_params)
+
+    if package_validation.valid?
+      package_response = packages.create rm_api_package
+
+      # Instead of trying to merge objects together, lets just request
+      # the newly created package from RM API and render that.
```
cowboyd: A good strategy, but should be handled by the model layer. In other words, the packages api should do the re-request if necessary.

Our move is gradually towards decoupling `mod-kb-ebsco` resources from flexirest resources (as in the case of `CustomLabel`) so this might be an option here.

spec/requests/packages_spec.rb
```diff
@@ -1061,4 +1061,125 @@
       end
     end
   end
+
+  describe 'creating a custom package' do
+    let(:update_headers) do
```
cowboyd, line 1140: Ruby convention is that imperative blocks are wrapped in `do`, side-effect free blocks, like this one are delimited by `{}`

app/controllers/packages_controller.rb
```diff
@@ -3,7 +3,7 @@
 class PackagesController < ApplicationController
   before_action :set_package, only: %i[show update customer_resources]
 
-  deserializable_resource :package, only: :update,
+  deserializable_resource :package, only: %i[create update],
```
cowboyd: `[:create, :update]` feels more straightforward. Is this the new rails convention?

app/controllers/packages_controller.rb
sadatay: 👍 gooood move

sadatay: although what if it doesn't exist?

app/controllers/packages_controller.rb
sadatay: if rubocop didn't catch this we gotta take its badge & gun

app/controllers/packages_controller.rb
Robdel12: Without `%i` the build fails due to rubocop.

app/controllers/packages_controller.rb
Robdel12: That's a good question. I could totally see an issue where we request the newly created package and the API doesn't return it. Worst case the package is still created and you get an empty response (instead of a full package)

Robdel12: Oh that's interesting. How would I go about doing this in the model? In order to return anything from this `POST` I have to request the package. This is because the only thing RM API responds with is `{ packageId: 11111 }`, which isn't great & makes the serializer fall over. 

So would the model be able to handle that? My mind is ready to be learned! :D

cowboyd: So by model layer here, I don't mean the RMAPI resource, I mean the controller's model. Right now, they are the same thing, and we are feeling a lot of pain because of it.

We want an abstract model layer whose primary allegiance is to capturing our app's REST API. In that world, the "create" method of the model, would decompose into _two_ requests of the RMAPI resource. If we can decouple our model in this way, then we can later ablate Flexirest without having to change the controller or serialization layers at all.
For example, we could move the current `Package` class into `RMAPIPackage`, and then make the `Package` model look like:

```ruby
class Package
  def self.create(attributes)
    package_response = rmapi_packages.create attributes
    rmapi_package = rmapi_packages.find(rmapi_package_id_for package_response)
    return new_from_rmapi_package rm_api_package
  end
end
```

In general, our life is going to be a lot easier if we don't account for RMAPI in our controllers at all. I don't know if this is possible, but I would time box it for a couple hours, and then proceed if it isn't feasible after that investment.

app/controllers/application_controller.rb
```diff
@@ -45,8 +45,12 @@ def catch_flexirest_exceptions
...
          Flexirest::HTTPNotFoundClientException => e
 
-    errors_hash = e.result.Errors.to_a.map do |err|
-      { "title": err.to_hash['Message'] }
+    if e.result.respond_to? :Errors
+      errors_hash = e.result.Errors.to_a.map do |err|
+        { "title": err.to_hash['Message'] }
+      end
+    else
+      errors_hash = { "title": e.result.to_hash['errors']&.items&.first&.message }
```
wwilsman: Should we just repeat what we have above, but with the different casing?

i.e.

```ruby
errors_hash = e.result.to_hash['errors'].items.map do |err|
  { "title": err.to_hash['message'] }
end
```

carolegodfrey (approved):
LGTM! - Helped having this as reference for Custom Resource work

PR #103, opened 2018-04-13
Delete endpoint for custom titles

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-294, users should be able to delete custom titles associated with custom packages, this is an endpoint for supporting the delete operation in backend.

## Approach
- Add a destroy method in the controller, which validates that it is a custom title which can be deleted and then sends a PUT request to RM API with "isSelected" set to false for this particular title.
- Sample request can be : DELETE /eholdings/resources/XX-YY-ZZ(vendorId-packageId-TitleId)
- Expected Response if title has been deleted successfully : 204 No Content
- Expected Response if its not a custom title and cannot be deleted: 400 Bad Request with detailed error message "Resource cannot be deleted"
- Expected Response if user tries to delete a title that does not exist: 404 Not Found with detailed error message: "Title not found"
- Associated unit tests

## Screenshots
delete_title

### DIFF
--- a/ModuleDescriptor.json
+++ b/ModuleDescriptor.json
@@ -39,7 +39,7 @@
           "pathPattern": "/eholdings/titles*"
         },
         {
-          "methods": [ "GET", "PUT" ],
+          "methods": [ "GET", "PUT", "DELETE" ],
           "pathPattern": "/eholdings/resources*"
         },
         {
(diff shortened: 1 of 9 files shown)


carolegodfrey (approved):
LGTM

PR #104, opened 2018-04-13
Delete custom/managed title

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-288, user should be able to delete a custom/managed title that is associated with a custom package.

## Approach
- Tweak code in custom resource delete param validation to check if title belongs to a custom package and if it does, support delete functionality with call to RM API setting "isSelected" for this specific title to false. 
- Modify tests accordingly.

### DIFF
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -29,7 +29,7 @@ class ResourcesController < ApplicationController
 
   def destroy
     resource_validation =
-      Validation::ResourceDestroyParameters.new(@resource)
+      Validation::ResourceDestroyParameters.new(@resource.customerResourcesList)
 
     if resource_validation.valid?
       @resource.delete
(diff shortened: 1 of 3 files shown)


cherewaty (approved):

app/controllers/resources_controller.rb
```diff
@@ -29,7 +29,7 @@ def update
 
   def destroy
     resource_validation =
-      Validation::ResourceDestroyParameters.new(@resource)
+      Validation::ResourceDestroyParameters.new(@resource.customerResourcesList)
```
sadatay, line 32: i'm unsure why we narrowed the scope of what we're passing here.  on the one hand we don't need the whole title record, i suppose.  but it's also already pretty confusing what makes a resource a 'resource' and it might be better off if we were consistent in the controllers about what it is we're acting on.  definitely not a blocker... odds are someone approaching this for the first time is going to have bigger questions than this.

sadatay (approved):

PR #105, opened 2018-04-16
UIEH-293 Enable more fields for custom title editing

cherewaty:
## Purpose
Set up editing for more editable fields on "custom titles" (custom resources).

https://issues.folio.org/browse/UIEH-293

## Approach
Makes editable:
- `isPeerReviewed`
- `publisherName`
- `edition`
- `description`
- `url`

Does not handle the two remaining array fields, `identifiers` and `contributors`.

## Next Steps
Does not attempt to do any validation. The five fields mentioned are not editable for a managed resource, so they should reject changes there. Until we get to that, we at least know the UI being built won't attempt to make those changes to a managed resource.

### DIFF
--- a/app/models/resource.rb
+++ b/app/models/resource.rb
@@ -110,7 +110,12 @@ class Resource < RmApiResource
       customEmbargoPeriod: resource_attributes[:customEmbargoPeriod],
       coverageStatement: resource_attributes[:coverageStatement],
       titleName: attributes[:titleName],
-      pubType: attributes[:pubType]
+      pubType: attributes[:pubType],
+      isPeerReviewed: attributes[:isPeerReviewed],
+      publisherName: attributes[:publisherName],
+      edition: attributes[:edition],
+      description: attributes[:description],
+      url: resource_attributes[:url]
     )
     refresh!
   end
@@ -158,7 +163,11 @@ class Resource < RmApiResource
   def update_fields
     to_hash.with_indifferent_access.slice(
       :titleName,
-      :pubType
+      :pubType,
+      :isPeerReviewed,
+      :publisherName,
+      :edition,
+      :description
     )
   end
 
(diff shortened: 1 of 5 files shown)


app/serializable/serializable_resource.rb
```diff
@@ -14,7 +14,11 @@ class SerializableResource < SerializableJSONAPIResource
              :isPeerReviewed,
              :isTitleCustom,
              :publisherName,
-             :titleId
+             :titleId,
+             :isPeerReviewed,
+             :publisherName,
```
sadatay: several of these are dupes - `publisherName` and `description` on first glance

cherewaty: Oops, fixed.

app/models/resource.rb
```diff
@@ -169,7 +178,7 @@ def resource_update_fields
       :customCoverageList,
       :customEmbargoPeriod,
       :coverageStatement,
-      :titleName
+      :url
```
sadatay, line 181: man flexirest is pushing us into some weird patterns :/ .  `update_fields` and `resource_update_fields`?

spec/requests/custom_resources_spec.rb
```diff
@@ -211,6 +211,85 @@
...
+      it 'has peer review status' do
+        expect(json.data.attributes.isPeerReviewed).to be true
+      end
+
+      it 'has publication type' do
+        expect(json.data.attributes.publicationType).to eq('Newspaper')
+      end
+
+      it 'has publisher name' do
+        expect(json.data.attributes.publisherName).to eq('Frontside Newspapers')
```
sadatay, line 278: 📰 👍

sadatay (changes requested):

sadatay (approved):

sadatay:
>Does not attempt to do any validation.

my only remaining comments were gonna be about this, but if we recognize it as such i think we're ok - given that we need to rethink models/validation anyway imo

app/models/resource.rb
cherewaty: There might be a more conventional way to deal with that, but that's what works for now.

PR #106, opened 2018-04-16
UIEH-321 Fix RM API package edit rejection

cherewaty:
## Purpose
https://issues.folio.org/browse/UIEH-321

When we send `contentType` or `packageName` in the request to `PUT` a managed package, RM API rejects the request completely.

Managed resources surprisingly don't have a problem with accepting uneditable attributes - they're just ignored.

## Approach
Don't send `contentType` or `packageName` changes to managed packages.

### DIFF
--- a/app/models/package.rb
+++ b/app/models/package.rb
@@ -117,13 +117,22 @@ class Package < RmApiResource
   end
 
   def update_fields
-    to_hash.with_indifferent_access.slice(
-      :isSelected,
-      :allowEbscoToAddTitles,
-      :visibilityData,
-      :customCoverage,
-      :packageName,
-      :contentType
-    )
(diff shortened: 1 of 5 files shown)


app/models/package.rb
```diff
@@ -117,13 +117,22 @@ def merge_fields(new_values)
...
+        :contentType
+      )
+    else
+      to_hash.with_indifferent_access.slice(
+        :isSelected,
+        :allowEbscoToAddTitles,
+        :visibilityData,
+        :customCoverage
+      )
+    end
```
cherewaty, line 136: @sadatay if you can do this more elegantly, feel free to add another commit to this PR.

sadatay (approved):

PR #108, opened 2018-04-18
DELETE endpoint for custom packages - UIEH-239

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-329, users should be able to delete a custom package. This PR implements functionality to support that.

## Approach
- Implement DELETE call, route it to destroy in package controller, validate if it is a custom package.
- If it is a custom package, send PUT request to RM API with "isSelected"=false for the corresponding vendor id and package id; and if package is deleted successfully, we respond with a `204 No Content`
- If it is not a custom package, then give a response of `400 - stating that the package cannot be deleted`
- If they attempt to delete a package that's already deleted, give a `404 - Package not found` as response.
- Add associated unit tests.
- Sample request: `DELETE /eholdings/packages/123355-2848225`
- Sample response: `204 No Content`

## Screenshots
custom_package_delete

### DIFF
--- a/ModuleDescriptor.json
+++ b/ModuleDescriptor.json
@@ -31,7 +31,7 @@
           "pathPattern": "/eholdings/providers*"
         },
         {
-          "methods": [ "GET", "PUT" ],
+          "methods": [ "GET", "PUT", "DELETE" ],
           "pathPattern": "/eholdings/packages*"
         },
         {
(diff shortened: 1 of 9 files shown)


carolegodfrey (approved):
Looks Great Sobha - just a minor comment on PR - if you add UIEH-239 to PR title it will be linked in Jira

sduvvuri-ebsco:
Thanks for your feedback Carole...added it to title.

PR #111, opened 2018-04-23
Update contributors list in PUT request for custom title

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-319, users should be able to edit "contributors" for a custom title. This implementation supports that.

## Approach
- Allow "contributors" to be passed to mod-kb-ebsco
- De-serialize it to "contributorsList" for RM API and send it in the PUT call
- Serialize the response from RM API to "contributors" and pass it in response
- Depend on RM API to validate supported contributor types
- Associated unit tests

#### TODOS and Open Questions
- [x] We have to do something similar to support the same field in POST request while creating a custom title. Is that a separate user story? - Increased the scope of this story to include the same for POST, will be a separate PR though.

## Screenshots
contributors_custom_title

### DIFF
--- a/spec/requests/custom_resources_spec.rb
+++ b/spec/requests/custom_resources_spec.rb
@@ -180,7 +180,7 @@ RSpec.describe 'Custom Resources', type: :request do
       end
     end
 
-    describe 'changing the embargo period' do
+    describe 'changing the coverage statement' do
       let(:params) do
         {
           "data": {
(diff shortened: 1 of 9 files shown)


spec/requests/custom_resources_spec.rb
```diff
@@ -235,6 +316,16 @@
                   "endCoverage": '2004-01-01'
                 }
               ],
+              "contributors": [
```
carolegodfrey, line 319: Looks great Sobha! - just one very minor thing - should we check for contributors in response body in this test as well?

sduvvuri-ebsco: Sure, done with latest commit. thanks Carole!

carolegodfrey (approved):

PR #112, opened 2018-04-24
Add contributors while creating a custom title - POST - UIEH-319

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-319, users should be able to add a list of contributors while creating a custom title.

## Approach
- Add contributors to allowed param validations
- Pass it through to RM API using resource create_resource model
- RM API returns contributor type values in lower-case, serialize them for the first character to be upper case for consistency in mod-kb-ebsco
- Associated unit tests

## Screenshots
post_contributor

### DIFF
--- a/app/controllers/validation/resource_create_parameters.rb
+++ b/app/controllers/validation/resource_create_parameters.rb
@@ -8,7 +8,7 @@ module Validation
 
     attr_accessor :titleName, :pubType, :packageId, :publisherName,
                   :isPeerReviewed, :edition, :description, :url,
-                  :customCoverageList,
+                  :customCoverageList, :contributorsList,
                   :embargoUnit, :embargoValue, :coverageStatement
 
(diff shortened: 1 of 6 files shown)


Robdel12 (approved):
This looks awesome! Love the gif, it helps me understand exactly the shape of data the API wants & how it works. 

I see the error response is perfect in the gif, but could we write a test for that? Just to make sure it doesn't break without us knowing

sduvvuri-ebsco:
@Robdel12 : Did you mean checking for the specific error message because a test testing the error code 400 is already in place. Please clarify. thanks!

Robdel12:
@sduvvuri-ebsco correct! Basically assert the payload comes back a "contributor type must be of type..." message

sduvvuri-ebsco:
thanks for your feedback @Robdel12! Done!

PR #113, opened 2018-04-24
Hotfix for customer resource updates

sadatay:
This is a quick fix for the broken edits we saw shortly before the demo today.  While `packageId` is a required field in the json payload for resource creation, it will break resource updates by overriding the existing packageId on the model - which is a computed property.  We're currently investigating why this didn't break in the tests, and that (along with hopefully a more elegant solution than the very un-DRY fix you see here) will be coming in a later PR.  For now, this is needed to unblock updates on resources wholesale.  It's currently running through telepresence on my box and at some point i'll have to close my laptop 😀

### DIFF
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -28,10 +28,10 @@ class ResourcesController < ApplicationController
 
   def update
     resource_validation =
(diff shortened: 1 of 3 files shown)


cherewaty (approved):
:shipit:

PR #114, opened 2018-04-25
Use okapi gem from rubygems.org

cowboyd:
## Purpose

When we started out, the okapi gem was not released to rubygems.org, so we linked directly to the git repostitory. It's now stable, and relased for awhile now, so we shouldbe consuming it just like any
other ruby app.

## Approach
Lock in to any of the 2.x releases. Before 2.0, this was actually a completey unrelated gem


Robdel12 (approved):
NICE 🎉 this looks good

sadatay (approved):
yezzir

PR #115, opened 2018-04-25
lazily fetch package in packages controller.

cowboyd:
## Purpose
There is currently a `before_action` that eagerly fetches the package
on an explicit set of requests (show, update, destroy,
resources). This has a couple of disadvantages.

1. The list of requests has to be manually curated, and if at some
point, the package isn't needed for one of the requests, it could
still be doing the work of fetching it unecessarily.
2. It uses rails magic `before_action :set_package` where none is
really necessary.

## Approach
Instead of magic, just use a regular ruby method that instantiates the
package lazily. This has the advantage of being normal Ruby, and also,
if only getting called when someone actually needs the package.

### DIFF
--- a/app/controllers/packages_controller.rb
+++ b/app/controllers/packages_controller.rb
@@ -1,8 +1,6 @@
 # frozen_string_literal: true
 
 class PackagesController < ApplicationController
-  before_action :set_package, only: %i[show update destroy resources]
-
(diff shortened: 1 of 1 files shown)


sadatay (approved):
this is a better approach - definitely a step in the right direction.

PR #109, opened 2018-04-18
UIEH-289 -Addition of Resource POST request to Create Custom Resources

carolegodfrey:
## Purpose
[UIEH-289](https://issues.folio.org/browse/UIEH-289)
Adds a POST method to `Resources` endpoint to support creating custom titles as part of a custom package. 

## Approach
- Sample request URL is of the following format:
`POST https://okapi.frontside.io/eholdings/resources`
- Sample request body with minimally required parameters is as follows:
`{"data": {"type": "resources","attributes": {"name": "Carole Custom Title Test Name", "publicationType": "Book", "packageId": 2843712}}}`
-  Followed similar pattern as in the following PR for creating custom packages https://github.com/folio-org/mod-kb-ebsco/pull/101. 
- Validates minimally required fields (name, publicationType and packageId) are present in request payload, otherwise response with 422
- `name` is maximum 400 characters
- `publicationType` is mapped to available RM API publication types or 'unspecified' if not found
- `packageId` is an integer
- Validates optional fields (if present) 
- `publisherName` is maximum 250 characters
- `isPeerReviewed` is true/false
- there is a current limitation in RM API - Post request expects `peerReviewed` instead of `isPeerReviewed` - this is inconsistent with with PUT request. Added temporary code to pass peerReviewed until this issue is resolved. Mod-kb-ebsco post request will continue to accept and return isPeerReviewed
- `edition` is a maximum 250 characters
- `description` is a maximum 1500 characters
- `url` is a maximum of 600 characters and starts with either http:// or https://
- `coverageStatement` is maxium 250 characters
- Added a common Provider `custom_provider_id` method to get the unique customer specific provider id.  This specific provider supports adding custom packages and custom resources. 
- Added a common Resource `create_resource` method  to interact with RM API to create custom resource
- A request is made to RM API to create a custom title in context of requested package id and vendor id (note - vendor id referenced as provider id from mod-kb-ebsco)
`POST https://api.ebsco.io/rm/rmaccounts/apidvcorp/vendors/123355/packages/2843712/titles`
`{
  "titleName": "Carole Custom Title Test Name",
  "pubType": "Book"
}`
- RM API POST request responds with a body containing newly created `titleId` if successful, for example
- `{"titleId": 17094867 }`
- An additional RM API GET request is made to retrieve newly created title  field 
`https://api.ebsco.io/rm/rmaccounts/apidvcorp/vendors/123355/packages/2843712/titles/17094867 `
[...]

PR #116, opened 2018-04-26
UIEH-343 Correct isPeerReviewed mapping on Resource Create

carolegodfrey:
## Purpose
While testing changes for custom resources create https://github.com/folio-org/mod-kb-ebsco/pull/109 an issue was encountered in RM API (POST request was accepting peerReviewed instead of isPeerReviewed and was inconsistent with PUT request). PR made adjustments in mod-kb-ebsco to handle this inconsistency. Since this issue has been corrected, this PR removes the temporary change.

## Approach
Removed temporary code which passed peerReviewed attribute instead of isPeerReviewed. Also adjusted unit test and re-recorded test cassette. (note - needed a new package since earlier referenced package was deleted)

### DIFF
--- a/app/models/resource.rb
+++ b/app/models/resource.rb
@@ -132,12 +132,8 @@ class Resource < RmApiResource
 
   def self.create_resource(params)
     package_id = params['packageId']
-    create_params = params.to_hash.except('packageId', 'isPeerReviewed')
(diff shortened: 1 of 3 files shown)


sduvvuri-ebsco (approved):
Looks good!

PR #117, opened 2018-04-27
Add identifiers for POST and PUT requests of custom titles - UIEH-318

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-318, users should be able to add identifiers while creating and updating custom titles. This PR addresses that issue. 

## Approach
- Serialize "identifiers" in mod-kb-ebsco to "identifiersList" in RM API
- An identifier can consist of "id", "type" and "subtype" - validate them and then pass them to RM API
- Sample POST request: `http://0.0.0.0:8081/eholdings/resources/`
- Sample POST request body containing identifiers: 
`{
 "data": {
  "type":"resources",
  "attributes": {
    "identifiers": [{"id": "12345","type":"ISBN","subtype":"Print"}}, {"id":12347,"type":"ISSN","subtype":"Online"}],
    "name": "POST test title",
    "publicationType": "Book",
	"packageId": 2850417
  }
 }
}`
Sample PUT request: PUT `http://0.0.0.0:8081/eholdings/resources/123355-2850417-17155900`
Sample PUT request body: 
`{
 "data": {
  "type":"resources",
  "attributes": {
    "identifiers": [{"id": "12345","type":"ISBN","subtype":"Print"}}, {"id":12347,"type":"ISSN","subtype":"Online"}],
    "name": "POST test title",
    "publicationType": "Book",
	"isSelected": true
  }
 }
}`
- Added associated unit tests

#### TODOS and Open Questions
- [ ] While deserializing, we are de-serializing only those types and sub-types that we are going to support for ui-eholdings, where as while serializing, we support types and sub-types that are not needed by ui-eholdings. I left the serialization as it was, because it is possible for whatever reasons that RM API could provide other types or sub-types where as I modified de-serialization because ui-eholdings cannot provide other types and sub-types. Should we keep them consistent? If so, shall we support all types and subtypes supported by RM API or only those that ui-eholdings needs?

## Screenshots
identifiers

### DIFF
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -89,6 +89,9 @@ class ResourcesController < ApplicationController
         contributorsList: [
           %i[type contributor]
         ],
+        identifiersList: [
+          %i[id subtype type]
+        ],
         customEmbargoPeriod: %i[
           embargoUnit
           embargoValue
@@ -117,6 +120,9 @@ class ResourcesController < ApplicationController
         ],
         contributorsList: [
           %i[type contributor]
+        ],
+        identifiersList: [
+          %i[id subtype type]
         ],
         customEmbargoPeriod: %i[
           embargoUnit
--- a/app/controllers/validation/resource_create_parameters.rb
+++ b/app/controllers/validation/resource_create_parameters.rb
@@ -8,7 +8,7 @@ module Validation
 
     attr_accessor :titleName, :pubType, :packageId, :publisherName,
                   :isPeerReviewed, :edition, :description, :url,
-                  :customCoverageList, :contributorsList,
+                  :customCoverageList, :contributorsList, :identifiersList,
                   :embargoUnit, :embargoValue, :coverageStatement
 
     validates :titleName, presence: true, length: { maximum: 400 }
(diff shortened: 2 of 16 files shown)


carolegodfrey (commented):
Looks really great @sduvvuri-ebsco 
Just checking -  is there some place where we are checking  that the identifiers update is only being applied for custom resources? Also I like keeping the type and subtype as separate attributes - but there was discussion around a single string passed from ui (containing both type and subtype). Not sure what was decide for what the ui would supply

sduvvuri-ebsco:
Thanks @carolegodfrey for reviewing. Its a very good point that identifiers could be edited only for custom resources. I totally missed that check. I will add it in. Will co-ordinate with @cherewaty on how UI is planning to send params. Did you get a chance to look at my open question? Thanks!

carolegodfrey:
I think that serialization and de-serialization is good as you have it. It is inconsistent (we only allow setting 4 specific types for custom titles for the de-serialize) - however for the serialize - I think we should show whatever rm api supports. Probably best to confirm this understanding with Khalilah though

cherewaty (approved):
Looks good 👍 This is a little different than my initial plan for how to pass the values between the UI and `mod-kb-ebsco`, but this is better, so let's go with this implementation.

sduvvuri-ebsco:
@carolegodfrey : Took care of being able to update identifiers only if it is a custom title and for the creation part, its being used only for custom titles as we discussed. Thanks!

carolegodfrey (approved):
Looks good!

PR #118, opened 2018-04-30
Accept all possible identifier types and subtypes that RM API has

cherewaty:
## Purpose
Updates to some resources return an error from `mod-kb-ebsco`.

Example: http://localhost:3000/eholdings/resources/18-2321-1239421
returns `Invalid Identifier type` and `Invalid Identifier subtype`.

The resource listed has identifier type "BHM" and subtype "Empty" from RM API. `PUT` requests to change anything about that resource will include that type and subtype.

## Approach
`mod-kb-ebsco` needs to be able to accept all the types/subtypes that RM API lists, not just ISSN/ISBN.


sduvvuri-ebsco (approved):
Cool, I had an open question of whether this could be a possibility on my PR and glad to know that it is a possibility :) Thanks for fixing it!

PR #119, opened 2018-05-01
Serialize Edition in titles -- Needed for UIEH-253

sduvvuri-ebsco:
## Purpose
per https://issues.folio.org/browse/UIEH-253, users should be able to edit edition field and view the new value on detailed record. From UI, to update the value, we use the /resources route in mod-kb-ebsco where as to view, we use the /titles route which does not serialize edition in its response.

## Approach
- Add edition in serializer to get it as part of response in mod-kb
- Add associated unit tests

## TODO and Open Questions
- Maybe, eventually we want to validate most fields of a resource while updating just like we do while creating,  I am not doing it in this PR because it needs more thought around what fields can be updated.

## Screenshots
edition_custom_title

### DIFF
--- a/app/controllers/validation/resource_update_parameters.rb
+++ b/app/controllers/validation/resource_update_parameters.rb
@@ -7,7 +7,10 @@ module Validation
     include ActiveModel::Validations
 
     attr_accessor :isSelected, :isHidden, :customCoverageList, :contributorsList,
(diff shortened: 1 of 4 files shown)


carolegodfrey (approved):
LGTM Sobha!

sduvvuri-ebsco:
Thanks @carolegodfrey

PR #120, opened 2018-05-02
Add title create route 

cherewaty:
## Purpose
EBSCO's RM API does not allow the creation of custom titles without creating a resource linking that title to a custom package. To make `mod-kb-ebsco`'s API more predictable, we'd like to establish a `POST /titles` endpoint.

## Approach
To work around RM API's relationship limitation, we can require titles to be created in conjunction with a resource, like:
```
{
  'data' => {
    'type' => 'titles',
    'attributes' => {
      'name' => 'New Title Testing',
      'publicationType' => 'Book'
    }
  },
  'included' => [
    {
      'type' => 'resources',
      'attributes' => {
        'packageId' => '2845504'
      }
    }
  ]
}
```

### DIFF
--- a/app/controllers/titles_controller.rb
+++ b/app/controllers/titles_controller.rb
@@ -54,7 +54,7 @@ class TitlesController < ApplicationController
     json = JSON.parse request.body.read
 
     unless json['included']&.first
-      raise ActionController::BadRequest 'Missing resource'
+      raise ActionController::BadRequest, 'Missing resource'
     end
 
     resource_params = json['data']['attributes'].merge(
--- a/app/controllers/titles_controller.rb
+++ b/app/controllers/titles_controller.rb
@@ -15,6 +15,22 @@ class TitlesController < ApplicationController
            meta: { totalResults: @titles.totalResults }
   end
 
+  def create
+    # RM API creates titles as a side-effect of creating resources, so
+    # we validate against a resource using the combined payload of the
+    # new title and an included resource with a packageId
+    resource_validation =
+      Validation::ResourceCreateParameters.new(combined_resource)
+
+    if resource_validation.valid?
+      @title = titles.create_title(combined_resource)
+      render jsonapi: @title
+    else
+      render jsonapi_errors: resource_validation.errors,
+             status: :unprocessable_entity
+    end
+  end
+
   def show
     render jsonapi: @title, include: params[:include]
   end
@@ -32,5 +48,22 @@ class TitlesController < ApplicationController
 
   def titles
     Title.configure(config)
+  end
+
+  def combined_resource # rubocop:disable Metrics/AbcSize
+    json = JSON.parse request.body.read
+
+    unless json['included']&.first
+      raise ActionController::BadRequest 'Missing resource'
+    end
+
+    resource_params = json['data']['attributes'].merge(
+      json['included'].first['attributes']
+    )
+
+    DeserializableResource.call(
+      'type' => 'resources',
+      'attributes' => resource_params
+    )
   end
 end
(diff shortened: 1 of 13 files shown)


wwilsman (approved):
Since I was the one who did most of this, not sure if my approval counts.

But to elaborate on what was done, I basically copied the logic from `ResourceController#create` and `Resource.create_resource` into the respective title controller and model. (Also copied the tests and made adjustments).

The only real addition is the `combined_resources` method in the title controller, which is needed because the JSON API parser has a hard requirement for `data` being the only top-level key in the incoming payload.

app/controllers/titles_controller.rb
```diff
@@ -15,6 +15,22 @@ def index
            meta: { totalResults: @titles.totalResults }
   end
 
+  def create
+    # RM API creates titles as a side-effect of creating resources, so
+    # we validate against a resource using the combined payload of the
+    # new title and an included resource with a packageId
+    resource_validation =
+      Validation::ResourceCreateParameters.new(combined_resource)
```
sadatay, line 23: so `ResourceCreateParameters` prior to this encompassed the title-level and resource-level attributes flattened into "our" concept of a resource?  why not split the validator in two now since we're splitting the concept (for updates, anyway)

cowboyd: I think this is the way to go, but let's tackle that work, when we change titles to use the repository pattern.

cowboyd (approved):

app/controllers/titles_controller.rb
```diff
@@ -33,4 +49,21 @@ def set_title
   def titles
     Title.configure(config)
   end
+
+  def combined_resource
+    json = JSON.parse request.body.read
```
sadatay, line 63:
```ruby
 rescue JSON::ParserError, NoMethodError
    error = {
      title: 'Invalid JSON',
      detail: 'The provided JSON payload could not be parsed'
    }

    render jsonapi_errors: [error],
           status: :unprocessable_entity
```

sadatay: and i remember you mentioned something about string keys, and you weren't sure why it was being enforced.  usually here we'd be dealing with `ActionController::Parameters` which implement "indifferent access" (string keys and symbol keys resolve identically. 

to roll your own:
```irb
response = HTTP.get('https://jsonplaceholder.typicode.com/posts/1')
response_in_business_casual = JSON.parse(response.body.to_s).deep_symbolize_keys!
response_in_sunday_best = Map response_in_business_casual
```

app/controllers/titles_controller.rb
```diff
@@ -33,4 +49,21 @@ def set_title
...
+
+    unless json['included']&.first
```
sadatay, line 65: `pretzel_dot`++

app/models/title.rb
```diff
@@ -86,4 +86,19 @@ def resources
...
+  # title that was also created
+  def self.create_title(params)
+    package_id = params[:packageId]
+    create_params = params.to_hash.except(:packageId)
+    rm_api_create = { vendor_id: provider_id, package_id: package_id }.merge(create_params)
+    resource_response = Resource.configure(config).create rm_api_create
+    find resource_response[:titleId]
+  end
+
+  def self.provider_id
```
sadatay, line 101: i feel like we should pick a better name than `provider_id`, especially if we'll be mirroring it into other models like this.  this is (at least in `Provider` itself) a special per-tenant provider to which all custom packages are initially assigned (if not explicitly told otherwise during creation.  `provider_id` is vague.  it conveys very little about what it actually is.

app/controllers/titles_controller.rb
sadatay: 👍 for future reference between now and the date that is approaching, unless i explicitly mark something as "NO.  NO NO NO." you can safely assume i'm a) asking something out of curiosity or b) making notes so we don't forget to address things later on.  i'm totally fine with waiting.

app/controllers/titles_controller.rb
wwilsman: Added the `rescue` to the `#create` portion so we don't end up with a double `render`.

PR #122, opened 2018-05-07
Link Resources to Custom Packages

sadatay:
## Purpose
Addresses UIEH-287.  This is backend work to support the linking of
both managed and custom titles to custom packages, thereby creating
custom resources.  To support this, the UI will need to be able to
filter packages by 'custom' anywhere package collections are dealt
with.  The UI will also need to be able to fire off a request that
will associate a custom package with a title and yield a unique,
compound ID representing the new "custom resource".

## Approach
There's a pretty major refactor shoved in here too.  For all other
`filter` query parameters, RMAPI supports them and they are passed
through directly.  In this case we needed to pull large collections of
packages over the wire one page at a time and then handle the
filtering logic inside of mod-kb-ebsco. This is, of course,
inefficient, but unless you're hitting `/packages` directly (which
takes about a minute to return) it's pretty bearable if scoped to a
vendor or filtered even further with other query params (which are
supported in parallel).  The rather massive refactor afforded us the
ability to finely tune how data is persisted.  It also de-couples us
from several layers of the application that we've noticed slowing us
down for the past several months - namely Flexirest, whose AR inspired
ORM layer was increasingly strained by the complexity of RMAPI.  Here
we use a pattern somewhere between a data mapper and a domain-specific
repository.  This separates the storage of data from its manipulation,
leading to clean models (which now look more like Codex structures
than RMAPI structures), slimmer controllers, and an open door for
complex queries such as the one needed to populate the drop-down for
this feature.

For now this overhaul was kept to the package route, which is thus far
the only place we truly needed to break free Flexirest / ORM pattern.
There will be documentation provided showing the major changes
involved.

## Usage

The `filter[custom]=true` query param can be passed to package routes: `/packages`, `/vendors/:vendorId/packages`.  The former takes about a minute to complete, and the latter can be combined with a query to `/titles/:titleId/resources` to derive which custom packages are valid for resource assignment.  Invalid assignment is also protected against in the API layer before hitting RMAPI.

To create a custom resource given a title and a custom package, fire off a  `POST /resources` with a payload comprised of the two IDs:

```
{
  "data": {
    "type": "resources",
    "attributes": {
      "packageId": "19-2516",
      "titleId": 481188
    }
  }
}
```

This will associate the title with the vendor/package to create a fully qualified resource with ID `19-2516-481188`.  Success returns a `200` with the body of the new resource.  In the specific example given above, a validation error will be thrown, as `19-2516` is a managed package.

## Learning
_links and a doc coming soon_

## Screenshots

**Creating a new Resource**
screen shot 2018-05-06 at 6 58 15 pm

**A validation error response**
screen shot 2018-05-06 at 7 44 56 pm

**Filtering `/packages` to only those which are custom.  ~40 out of 10k**
screen shot 2018-05-06 at 11 16 14 pm

**Long-running request.  Mapped to 100 requests to RMAPI at 100 items per page**
screen shot 2018-05-06 at 11 16 31 pm

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -74,9 +74,6 @@ pipeline {
 
         sh "sudo /bin/bash -l -c '. /usr/share/rvm/scripts/rvm && bundle install'"
 
-        echo "Run Ruby linter - rubocop..."
-        sh "/bin/bash -l -c '. /usr/share/rvm/scripts/rvm && rubocop'"
-
         echo "Run unit tests..."
         sh "/bin/bash -l -c '. /usr/share/rvm/scripts/rvm && rake spec'"
       }
@@ -135,5 +132,3 @@ pipeline {
   }
 
 }
-
-


app/controllers/packages_controller.rb
```diff
@@ -1,70 +1,38 @@
 # frozen_string_literal: true
 
 class PackagesController < ApplicationController
+  # attr_accessor :package
+  # before_action :set_package, only: %i[show update destroy resources]
```
cowboyd: Let's remove dead code.

cowboyd (commented):
With a change of this magnitude, I think we need to smoke test it on sandbox first. Tests are passing, but need to make sure it all works against current deployed ui

cowboyd (approved):

PR #129, opened 2018-05-09
Expect composite package ids when creating titles

cherewaty:
## Purpose
Custom titles weren't being created, because `ui-eholdings` only knows about the composite `packageId` of the RM API vendor id + RM API package id.

## Approach
Deconstruct the composite id to just get the RM API package id.

Tests updated to only use the composite ids.


Robdel12 (approved):
🎉

PR #130, opened 2018-05-09
Do not POST empty arrays to RM API for custom titles

cherewaty:
## Purpose
If you `POST` to RM API to create a custom title, and `contributorsList` or `identifiersList` are empty arrays, it returns a 500 with error `Malformed request`.

So it was impossible to create custom titles in the front end without contributors or identifiers.

## Approach
Eliminate `contributorsList` or `identifiersList` from the request params if `POST`ing and they're empty.


carolegodfrey (approved):
LGTM!

PR #132, opened 2018-05-09
UIEH-135: Mask RM API encryption key when it is read from configuration endpoint

carolegodfrey:
[UIEH-135 - RM API Encryption Key - do not return key in GET response](https://issues.folio.org/browse/UIEH-135)

## Purpose
eHoldings application does not display the RM API encryption key in the UI. As was implemented in [UIEH-105](https://issues.folio.org/browse/UIEH-105). This follow on ticket updates mod-kb-ebsco GET endpoint to prevent the return of credentials. As it exists now, credentials can be obtained by observing network traffic or viewing text of the settings page. 

## Approach
As suggested in earlier ticket, return a masked string (********) 40 characters for RM API from configuration GET request

#### TODOS and Open Questions
- [ ] Longer term need to consider storage of key in database - should it be encrypted there?

## Screenshots
GET response from mod-kb-ebsco - 
mask-api-key
Settings page from which credentials can be observed.
settings-api-key

### DIFF
--- a/app/serializable/serializable_configuration.rb
+++ b/app/serializable/serializable_configuration.rb
@@ -3,5 +3,11 @@
 class SerializableConfiguration < SerializableJSONAPIResource
   type 'configurations'
 
-  attributes :api_key, :customer_id
+  attribute :customer_id
+
+  # The api key is masked in the payload returned from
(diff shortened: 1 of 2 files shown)


cherewaty (approved):
👍

PR #133, opened 2018-05-10
Publish API docs in CI

funkymalc:
## Purpose
Publish API documentation 

## Approach
Add Publish API doc step in Jenkinsfile

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -121,6 +121,22 @@ pipeline {
       }
     }
 
+    stage('Publish API docs') { 
+      when {
+        branch 'master'
+      }
+      steps { 
+        echo "Publishing API docs"
+        sh "python3 /usr/local/bin/generate_api_docs.py -r $env.project_name -v -o folio-api-docs"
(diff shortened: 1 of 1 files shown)


cowboyd:
Thanks @funkymalc! Will this pick up any `*.raml` file in the `/ramls` directory, or is that set by the config file that @dcrossleyau maintains?

dcrossleyau:
At this time, the config needs to explicitly list each raml file. Job when i get back to office to enhance that. https://github.com/folio-org/folio-org.github.io/blob/master/_data/api.yml

cowboyd:
That makes sense. 

So if we go ahead and merge this without any ramls, it won't break anything, and then once we add the top-level raml to repository, I can pass you the name of the file and we should be good to go?

dcrossleyau:
I was expecting you to add your initial skeleton raml branch first, then this Jenkinsfile PR. The api.yml  config is already deployed with "eholdings.raml" as the first.

dcrossleyau:
Any issues with the generation of API docs do not break the CI, just emit messages. So safe to merge in whatever order.

cowboyd (approved):

PR #135, opened 2018-05-16
UIEH-370 - Handle proxy setting at provider level

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-370, users should be able to view the current proxy setting as well as update selection at the provider level. This PR supports the same.

## Approach
- Add support for de-serializing and serializing proxy information in the GET and PUT calls for provider.
- Associated unit tests.

#### TODOS and Open Questions
- [ ] Do we want to deal with the additional complexity of "inherited" in backend or will UI be able to handle that? - A discussion in progress..

## Screenshots
proxy_provider

### DIFF
--- a/app/controllers/providers_controller.rb
+++ b/app/controllers/providers_controller.rb
@@ -60,7 +60,8 @@ class ProvidersController < ApplicationController
     params
       .require(:provider)
       .permit(
-        vendorToken: [:value]
+        vendorToken: [:value],
+        proxy: %i[id inherited]
       )
   end
 end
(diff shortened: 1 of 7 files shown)


carolegodfrey (approved):
These changes looks good to me @Sobha - just one question/comment - will we be referring to the proxy id <n> as none eventually and may want to always do a translation? Assume wouldn't address until other ticket is worked on

sduvvuri-ebsco:
yes Carole, its part of another ticket  - UIEH-264 and as part of that, we will deal with serializing and deserializing <n> to None.

PR #136, opened 2018-05-16
UIEH-376 : Proxy support at resource level

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-376, users should be able to view/update proxy setting at a resource(title-package) level irrespective of whether it is a managed title or a custom title.

## Approach
- Add proxy in serializers and deserializers and in update params of a resource.
- Associated unit tests

## Learning
- What I learnt while implementing this is that RM API, if current proxy setting is inherited, then for updating the setting, we need to pass "inherited" as false; else RM API will throw an error that inherited proxy setting cannot be updated. Where as when we do a GET, the state of inherited need not match. RM API gives the correct state of whether it is inherited or not irrespective of what we pass to it. Added support to get both proxy "id" as well as "inherited" from UI. 

## Screenshots
root_proxy_resource

### DIFF
--- a/spec/requests/resources_spec.rb
+++ b/spec/requests/resources_spec.rb
@@ -45,11 +45,12 @@ RSpec.describe 'Resources', type: :request do
         'url',
         'vendorId',
         'vendorName',
-        'visibilityData'
+        'visibilityData',
+        'proxy'
       )
     end
 
-    it "has a composite pacakge id of '{vendor_id}-{package_id}'" do
+    it "has a composite package id of '{vendor_id}-{package_id}'" do
       expect(attributes.packageId).to eq('22-1887786')
     end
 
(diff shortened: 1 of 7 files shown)


spec/requests/resources_spec.rb
```diff
@@ -387,6 +397,10 @@
                 'isSelected' => true,
                 'visibilityData' => {
                   'isHidden' => true
+                },
+                'proxy' => {
```
sadatay, line 401: nice, for some reason i was under the impression that the proxy payload and resource payload (both managed and custom) were disjunct, and we'd need to do two calls in the background.  if this is working i'm really glad we don't have do that

sadatay (approved):

PR #139, opened 2018-05-18
UIEH-371 UIEH-374 Root proxy refactor repository pattern

elrickvm:
## Purpose
The knowledge has the concept of Root Proxy that a user can set at the base level of the eholdings application in their Settings. The goal is to have a dropdown menu that will allow a user to select from all the available Proxy and then send request to update their Root Proxy from the available options in the list. 

Prior we were handling both requests: a) GET - to get all the the available Proxies and b) PUT - to update the root Proxy in one endpoint. That presented a challenge because we were getting a collection of available proxies but needed to return a request of just one proxy and not the collection. This is deemed a Singular Resource.

Separating the GET and PUT into two endpoints will allow the retrieval of the collection and the update of the singular resource to happen independently.

## Approach
  - Separate the concept of Proxy into two entities:
    - `Proxy-Types`
    - `Root-Proxy`
  - Create two endpoint for either entity:
    - `eholdings/proxy-types` 
         _availble requests_
      -  GET - will return all available proxies
    - `eholdings/root-proxy` 
         _availble requests_
      -  GET - will return currently set root-proxy
      -  PUT - will update the root-proxy

## Screenshots
  #### GET all proxy-types
  
screen shot 2018-05-17 at 9 33 07 pm

  ##### Setting a new root proxy
screen shot 2018-05-17 at 9 36 16 pm

  #### Error attempting to set a Root Proxy that is not in the available proxy-types list
screen shot 2018-05-24 at 8 00 38 pm

 #### Error attempting to set a Root Proxy with empty attributes
screen shot 2018-05-24 at 7 58 57 pm


app/models/proxy_type.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class ProxyType
+  include ActiveAttr::Model
+
+  attribute :id
+  attribute :name
+  attribute :url_mask
```
sduvvuri-ebsco, line 8: I see that we are not providing "selected" as an attribute here...is the idea that UI will make 2 calls, 1. to get the list and 2. to get the current root-proxy selection to be able to highlight the current selection in the drop-down?

app/repositories/root_proxies_repository.rb
```diff
@@ -0,0 +1,111 @@
...
+    proxy_id = attrs[:proxyTypeId]
+    request do
+      root_proxy_validation = Validation::RootProxyParameters.new(proxy_id, proxy_types_list)
+      fail ValidationError, root_proxy_validation unless root_proxy_validation.valid?
+
+      payload = attrs.to_hash.deep_symbolize_keys
+
+      payload[:proxy] = { id: payload.delete(:proxyTypeId) }
+
+      rmapi(:put, '', json: payload)
```
sduvvuri-ebsco: Per our discussion regarding packages refactor, we should add some error handling around the PUT such that if it fails for whatever reason, following GET should not be called?

app/serializable/serializable_proxy_type.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class SerializableProxyType < SerializableJSONAPIResource
+  type 'proxyType'
+
+  # Custom Label attributes
+  attributes :id,
+             :name,
+             :url_mask
```
sduvvuri-ebsco, line 9: "" as above regarding "selected"

app/repositories/root_proxies_repository.rb
```diff
@@ -0,0 +1,111 @@
...
+  def update!(attrs)
+    proxy_types_list = ProxyTypesRepository.new(config: @config).where!
+    proxy_id = attrs[:proxyTypeId]
+    request do
+      root_proxy_validation = Validation::RootProxyParameters.new(proxy_id, proxy_types_list)
+      fail ValidationError, root_proxy_validation unless root_proxy_validation.valid?
+
+      payload = attrs.to_hash.deep_symbolize_keys
+
+      payload[:proxy] = { id: payload.delete(:proxyTypeId) }
```
sduvvuri-ebsco: Here payload that we get from UI-eholdings will contain only proxy-related information, however, for RM API, proxies and custom-labels are handled by the same PUT call, so we need to make a GET, extract custom-labels and append it to payload before we pass it to RM API; else it would reset all custom-labels.

app/repositories/root_proxies_repository.rb
elrickvm: You are absolutely correct! on the error handling on PUT's. I was waiting to see the approach that was taking so i can copy that pattern in there.

app/repositories/root_proxies_repository.rb
elrickvm: Yep! We learned that lesson on Thursday evening. I will add the call to get the custom labels today.

app/models/proxy_type.rb
elrickvm: That was my impression from conversation with @sadatay and @cowboyd. I can go either way.

app/controllers/proxy_types_controller.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class ProxyTypesController < ApplicationController
+  def index
+    @result = proxy_types.where!
```
sadatay: it'd be cool to have a method on the repo interface called `all!` that just calls `where!` without params.  just an alias really, but it'd be more readable in situations like this.  definitely not in scope here, just a note for the future

app/controllers/root_proxies_controller.rb
```diff
@@ -1,43 +1,34 @@
 # frozen_string_literal: true
 
 class RootProxiesController < ApplicationController
-  def index
-    render jsonapi: root_proxies.all, include: params[:include]
+  deserializable_resource :root_proxy, only: [:update],
+                                       class: DeserializableRootProxy
+
+  def show
+    @result = root_proxy.find!
```
sadatay, line 8: doesn't this need an id?

app/repositories/packages_repository.rb
```diff
@@ -17,7 +17,6 @@ def initialize(config:)
   # CRUD Operations
   def find!(id)
     request do
-      # binding.pry
```
sadatay: ✂️

app/repositories/proxy_types_repository.rb
```diff
@@ -0,0 +1,83 @@
...
+  def initialize(config:)
+    @config = config
+    @base_url = "#{rmapi_url}/rm/rmaccounts/#{config.customer_id}/proxies"
+
+    @headers = {
+      'X-Api-Key': config.api_key,
+      'Content-Type': 'application/json',
+      'Accept': 'application/json'
+    }
+  end
```
sadatay: so, now that there are multiple repositories, there are a bunch of things that we should pull out and centralize with respect to all repo classes.  i'm not saying we should do all of that here, but if there's time maybe we could do something easy like yank these headers out into a superclass?

app/repositories/proxy_types_repository.rb
```diff
@@ -0,0 +1,83 @@
...
+        data: body.map { |hash| to_proxy_type hash },
+        status: status
+      )
+    end
+  end
+
+  private
+
+  # superclass for repository errors
+  class RepositoryError < StandardError; end
```
sadatay: these also would be easy - maybe even easier than the stuff in `initialize`.  they're class definitions, we can just put them in a separate file (or files) in the same namespace and they should be usable by all repos.  no inheritance or anything, just literal copy-paste

spec/requests/vendors_spec.rb
```diff
@@ -73,7 +73,6 @@
     let!(:json) { Map JSON.parse response.body }
 
     it 'gets associated package records' do
-      # binding.pry
```
sadatay, line 76: ✂️

app/controllers/root_proxies_controller.rb
elrickvm: Not sure. Because i think it just calls what is deemed the base_url and peels off the proxy

app/repositories/packages_repository.rb
elrickvm: 100%

app/controllers/root_proxies_controller.rb
cowboyd: No, because this is a singular resource

app/controllers/proxy_types_controller.rb
cowboyd: Let's do it! This makes total sense.

app/controllers/root_proxies_controller.rb
cowboyd: For the record, no. because this is a singular resource.

app/controllers/root_proxies_controller.rb
```diff
@@ -1,43 +1,43 @@
...
 
   def update
-    data_attributes = JSON.parse(request.body.read)['data']['attributes'] || {}
-    root_proxy_id = params[:id]
-
-    root_proxy_validation =
-      Validation::RootProxyParameters.new(data_attributes,
-                                          root_proxy_id,
-                                          root_proxies.all)
+    proxy_types_list = ProxyTypesRepository.new(config: config).all!
```
cowboyd, line 13: All of this work is the responsibility of the repository.

the controller method should look like

```ruby
def update
  @result = root_proxy.update! root_proxy_update_params
  render jsonapi: :no_content # 204 No Content or however it's done in rails
end
```

everything else, validation, error handling, pre-requisite reads, are all the responsibility of the repository. The controller is only concerned with HTTP.

cowboyd (changes requested):

app/repositories/root_proxies_repository.rb
```diff
@@ -32,7 +32,14 @@ def update!(attrs)
       # custom-labels along all the custom if any will be erased.
       # so we need to get the custom labels and post those along
       # with the root-proxy
+
+      _, custom_label_body = rmapi(:get, '')
```
cowboyd: Is this a bug? If the request fails, then it will continue processing.

app/controllers/root_proxies_controller.rb
elrickvm: @sadatay said that it was settle upon earlier that because of the side-effects that RMAPI does that we would return a body.

sadatay: ^ this.  if we want to be consistent with the other endpoints anyway.  the reasoning behind using a 204 depends on knowing with certainty that a resource + an update = current state.  we don't have that certainty elsewhere in rmapi so we return resource bodies on update everywhere else.  

wrt what is and isn't the responsibility of the repository, i do feel like we shouldn't be getting proxytypes here.  it appears to be needed for validation, and maybe this is a scenario where validation shouldn't be in the controller.

cowboyd (approved):

PR #140, opened 2018-05-18
FOLIO-1187 lint raml cop

dcrossleyau:
## Purpose
Add scripts/lint-raml-cop.sh to assist with local testing.
Add the shared raml-util git submodule, which will probably be needed soon. If not then it can be removed later.


dcrossleyau:
There is a bit of a Catch-22. When this shared raml is in-place then the generate-api-docs CI job can be further configured.

cowboyd (approved):

PR #141, opened 2018-05-22
UIEH-384 - Set proxy inherited=false until UI is implemented

sduvvuri-ebsco:
## Purpose
- At provider level and resource level, proxy is implemented to be accepted in PUT calls. Proxy has two attributes -> id and inherited. For RM API, we are still unclear on whether it needs the "inherited" attribute. We are discussing with Mohan about it. Until then and until UI is implemented, for a PUT, we make a GET call which gives "inherited=true" and when we are trying to pass the same back to RM API, it throws an error that "id cannot be present when inherited is true". For RM API, when inherited is true, proxy id has to be "".

## Approach
- Hard code "inherited" attribute of proxy to false before sending an update(PUT)
- Do not accept "inherited" as a param from UI; we might not need it
- Modify associated tests; some of the existing tests failed because they do not have "proxy" information; where as realistically, in a PUT request, we should get the complete payload; so added "proxy" to the payload of those tests. 

## Screenshots
proxy_inherited

### DIFF
--- a/app/controllers/providers_controller.rb
+++ b/app/controllers/providers_controller.rb
@@ -61,7 +61,7 @@ class ProvidersController < ApplicationController
       .require(:provider)
       .permit(
         vendorToken: [:value],
-        proxy: %i[id inherited]
+        proxy: [:id]
       )
   end
 end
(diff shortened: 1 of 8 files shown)


carolegodfrey (approved):
LGTM @sduvvuri-ebsco - thanks!

PR #142, opened 2018-05-22
Part of UIEH-362: Separate serializers at provider level

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-362, we should separate serializers in mod-kb based on action to not give incorrect data to ui-eholdings. 
Unfortunately, RM API provides different attributes for list vs. detail record and this fix ensures that we do not provide what RM API does not provide to us due to common serializer in mod-kb.

## Approach
- Investigate which attributes differ between a providers' list and provider detail record - its one attribute `proxy` at the detail record that is in addition to the attributes that are provided as part of the providers' list
- Separate provider serializers into `SerializableProviderList` and `SerializableProvider` -> which is the default. 
- Inherit  `SerializableProvider` from `SerializableProviderList` and add additional attributes corresponding to a detailed record.
- Update associated unit tests to ensure that list does not provide `proxy` as part of its attributes where as detailed provider record does.

## Learning
- Initially, I tried to name `SerializableProvider` as `SerializableProviderShowUpdate` based on action - but it fails here -> /.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/jsonapi-serializable-0.3.0/lib/jsonapi/serializable/relationship/dsl.rb:24:in `block (2 levels) in data' because it is looking for SerializablePackage when we include packages and if we were to provide another class, then we have to do something like `class: { Provider: SerializableProvidershowUpdate }` to specify that serializer and when we include packages, its looking for SerializablePackage by default and we cannot provide 2 separate serializers in one render method. 
- So, to get around the problem above, decided to keep one name default as Rails would do it and the other would be `Serializable*List` corresponding to a list; also thought this would be better when we refactor serializers at other levels.
- I am open to suggestions on the naming convention here.
 
## Screenshots
separate_provider_serializer

### DIFF
--- a/app/controllers/providers_controller.rb
+++ b/app/controllers/providers_controller.rb
@@ -14,11 +14,13 @@ class ProvidersController < ApplicationController
     )
 
     render jsonapi: @providers.vendors.to_a,
-           meta: { totalResults: @providers.totalResults }
+           meta: { totalResults: @providers.totalResults },
+           class: { Provider: SerializableProviderList }
   end
 
   def show
-    render jsonapi: @provider, include: params[:include]
+    render jsonapi: @provider,
+           include: params[:include]
   end
 
   def update
--- a/app/serializable/serializable_provider.rb
+++ b/app/serializable/serializable_provider.rb
@@ -1,23 +1,5 @@
 # frozen_string_literal: true
 
-class SerializableProvider < SerializableJSONAPIResource
-  type 'providers'
-
-  attribute :name do
-    @object.vendorName
-  end
-
-  attributes :packagesTotal,
-             :packagesSelected,
-             :proxy
-
-  attribute :providerToken do
-    @object.vendorToken
-  end
-
-  attribute :supportsCustomPackages do
-    @object.isCustomer
-  end
-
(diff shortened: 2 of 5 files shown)


wwilsman (approved):
This looks great to me!

app/controllers/providers_controller.rb
```diff
@@ -14,11 +14,13 @@ def index
     )
 
     render jsonapi: @providers.vendors.to_a,
-           meta: { totalResults: @providers.totalResults }
+           meta: { totalResults: @providers.totalResults },
+           class: { Provider: SerializableProviderList }
```
cowboyd, line 18: It might do with a brief comment here to show that we're using a different serializer here to work around a bug in RMAPI, and that we can use the normal serializer once that fix has been deployed. It would be a major bonus to have some way to reference the status of the bug so that someone coming along later.

sduvvuri-ebsco: Sure, sounds good. I will add comments both at provider level and package level when I work on resources. Thanks for your feedback @cowboyd !

PR #144, opened 2018-05-22
Part of UIEH-362 - Separate serializers at title level

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-362, we are supposed to separate serializers based on action. This PR creates that separation for titles.

## Approach
- Create 2 separate serializers `SerializableTitleList` and `SerializableTitle`
- `SerializableTitle` inherits from `SerializableTitleList` and adds those additional fields that are part of the detailed record but not part of list(a discrepancy in RM API)
- Additional fields are `description`, `edition`, `isPeerReviewed` and `contributors`
- Update unit tests accordingly to ensure that the above mentioned fields are included in a detailed record request but not included in list that is the result of a search.
- Added comments to explain why we need separation of serializers.

## Screenshots
separate_titles_serializer


wwilsman (approved):
🎉

PR #145, opened 2018-05-29
First pass at RAML for packages endpoint

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-377, we are supposed to hand-roll RAML to document the `/packages` endpoint. This is a first attempt to hand-roll some RAML and know if I am moving the right direction.

## Approach
- Following https://raml.org/developers/raml-200-tutorial, generate resourceTypes.raml, samples for collection and collection-item and use them in the raml file.

#### TODOS and Open Questions
- [ ] Haven't created/used any traits yet
- [ ] Do we need to do something special for `application/vnd.api+json`?

## Learning
https://raml.org/developers/raml-100-tutorial
https://raml.org/developers/raml-200-tutorial

## Screenshots

### DIFF
--- a/ramls/types/collection-item.raml
+++ b/ramls/types/collection-item.raml
@@ -1,0 +1,16 @@
+collection-item:
+  description: Entity representing a <<resourcePathName|!singularize>>
+  get:
+    description: |
+      Get the <<resourcePathName|!singularize>>
+      with <<resourcePathName|!singularize>>Id =
+      {<<resourcePathName|!singularize>>Id}
+    responses:
+      200:
+        body:
+          application/vnd.api+json:
+      404:
+        body:
+          application/vnd.api+json:
+            example: |
+              {"message": "<<resourcePathName|!singularize>> not found" }
--- a/ramls/eholdings.raml
+++ b/ramls/eholdings.raml
@@ -1,4 +1,4 @@
-#%RAML 0.8
+#%RAML 1.0
 title: "mod-kb-ebsco"
 baseUri: https://github.com/folio-org/mod-kb-ebsco
 version: v1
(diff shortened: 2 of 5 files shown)


dcrossleyau:
Please follow the [coding style](https://dev.folio.org/guidelines/contributing/#coding-style) (no tabs and consistent two-space indent).

Please comment-out the 'Publish API docs' in ./Jenkinsfile until the generate-api-docs CI job is ready to handle RAML-1.0 ([FOLIO-1253](https://issues.folio.org/browse/FOLIO-1253)).

sduvvuri-ebsco:
@dcrossleyau : thanks for your feedback! By commenting out that line in ./Jenkinsfile, the documentation will not be published to dev.folio.org, correct? Want to understand the impact of commenting that line out. Thanks!

dcrossleyau:
That is correct. As discussed in Slack, if you are not ready to merge yet, then just wait. I should be ready tomorrow to deploy my updated CI script that builds the docs. If you do want to merge ASAP, then it will fail at that point. But the merge to master would be successful.

ramls/types/collection-item.raml
```diff
@@ -0,0 +1,16 @@
+collection-item:
+  description: Entity representing a <<resourcePathName|!singularize>>
```
keathleydavidj, line 2: So where does this `resourcePathName` come from in these resourceType files? Does it just use the string in the URL that this gets attached to via `type` above?

sduvvuri-ebsco: thats my understanding but I am modifying this back to following 100 Tutorial, the PR for which I will push by EOD. I think this reusable way they suggest is purely for RESTful APIs where the data matches exactly between GET and POSTs, which is not our case. We'll see which way to proceed after we have both to compare.

dcrossleyau:
The raml2html generate-api-docs CI job is now upgraded to handle RAML-1.0 version files.

Please see notes at [FOLIO-1253](https://issues.folio.org/browse/FOLIO-1253) and [UIEH-277](https://issues.folio.org/browse/UIEH-277).

cowboyd (approved):
I saw that @dcrossleyau had installed a raml linter. Does this still pass that check with the spacing? Is there a way to run that manually to verify?

sduvvuri-ebsco:
This is one of the advanced approaches that we could model our RAML around if we wanted to but having done this, I figured that it assumes that an API is purely RESTful and with us passing through to RM API which is not completely RESTful(for example: the outcome of a GET request and payloads for POST do not match), so I went ahead with the more verbose rudimentary approach which is in this PR:  https://github.com/folio-org/mod-kb-ebsco/pull/148 followed by  https://github.com/folio-org/mod-kb-ebsco/pull/150; both of which have been merged to master and passed RAML lint.

dcrossleyau:
@cowboyd Yes the script can be run manually (#149). In mod-kb-ebsco do ./scripts/lint-raml-cop.sh

See other notes at [UIEH-277](https://issues.folio.org/browse/UIEH-277)

There are still some errors in master. It is important to attend to those. They can mask other issues. See other notes at https://dev.folio.org/guides/raml-cop/

PR #146, opened 2018-05-30
UIEH-361 Adjust error handling for PackagesRepository

sadatay:
## Purpose

In the push to get this out for the demo we botched validation from a
few angles with respect to the new pattern.  This broke almost (if not
all) edits on the package details page.

## Approach
The following work addresses this, while leaving some of the
underlying issues unaddressed for now.  Here's a summary of what was completed:

- Immediate fix for 361, which consists of a) raising an exception if
  a non 200-level response code is encountered anywhere in a
  repository operation.  This will prevent the false positives.  The
  actual 'contentType' error that was being seen is, for now,
  sidestepped by branching on the package's `custom` status inside the
  repo's `update` method.
  - Implementing point a) involved creating an RmapiResource
  superclass to hold some common functionatliy and then adjusting the
  `rmapi` method to a more generic `request` method that fails
  immediately on any HTTP error code.  This failure bubbles up and is
  caught by our other error handlers.
- Validation was pulled into the controller level, because research
  indicated that was the convention, and because it also makes more
  sense there (the way things are currently).  It's not exactly the
  repository's job to validate what it intends to persist.  It should
  only persist, and should only receive arguments ready for
  persistance imo.
- The serialization concerns about case clashing have been addressed
  as far as I'vde seen
- The tests were adjusted accordingly.

#### TODOS and Open Questions
- [ ] The fact that the tests were green when we last merged this gives me very little confidence in their current efficacy.  This should be of high concern.  The tests in this case impeded my changes rather than empowered them.  
- [ ] For now the data transformation is still inline.  I wanted to get an MVP out so I avoided pulling them out into (at the very least) separate methods for now or (ideally) some kind of data structure that represents communication between rmapi and mod-kb in both directions.  I'd like to circle back on that today if i get time, but if not we'll get to it another time.  
- [ ] There are several other TODOs in the code as well, but this covers the ones that are major enough and mature enough to warrant additional time allocated.

## Learning
- [some stuff about how validation works in laravel](https://laracasts.com/discuss/channels/laravel/repository-pattern-and-validation)
- [general overview of repository pattern](https://martinfowler.com/eaaCatalog/repository.html) which leaves some open questions up to the implementor(s) but otherwise is a nice summary

### DIFF
--- a/app/repositories/packages_repository.rb
+++ b/app/repositories/packages_repository.rb
@@ -1,162 +1,105 @@
 # frozen_string_literal: true
 
-class PackagesRepository
-  attr_reader :vendor_id, :package_id, :base_url, :headers
-
-  def initialize(config:)
-    @config = config
-    @base_url = "#{rmapi_url}/rm/rmaccounts/#{config.customer_id}"
-
-    @headers = {
-      'X-Api-Key': config.api_key,
-      'Content-Type': 'application/json',
-      'Accept': 'application/json'
-    }
-  end
-
-  # CRUD Operations
+class PackagesRepository < RmapiRepository
   def find!(id)
-    request do
-      # binding.pry
-      vendor_id, package_id = id.split('-')
-      status, body = rmapi(:get, "/vendors/#{vendor_id}/packages/#{package_id}")
-      Result.new(data: to_package(body), status: status, included: body[:included])
-    end
+    vendor_id, package_id = id.split('-')
+    status, body = request(:get, "/vendors/#{vendor_id}/packages/#{package_id}")
+    Result.new(data: to_package(body), status: status, included: body[:included])
   end
 
   def create!(attrs)
-    request do
-      payload = attrs.to_hash.deep_symbolize_keys
-      payload[:packageName] = payload.delete(:name)
-      package_validation = Validation::CustomPackageParameters.new(payload)
-      fail ValidationError, package_validation unless package_validation.valid?
+    payload = attrs.to_hash.deep_symbolize_keys
+    payload[:packageName] = payload.delete(:name)
 
-      content_type_enum = {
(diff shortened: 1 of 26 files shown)


app/repositories/packages_repository.rb
```diff
@@ -1,162 +1,105 @@
...
+      if params.dig(:filter, :custom)
+        # The 'custom' filter option is unique in that
+        # it is not passed through to RMAPI.
+        #
+        # TODO: remove mutation. params sent to rmapi should be completely
+        # mapped from scratch every time, not a munge of controller params
+        # (cowboyd 5/3/2018)
+        params[:filter].delete(:custom)
+
+        # All custom packages have the same vendor_id per
```
elrickvm, line 85: per what? Seems like it could be expanded upon a little.

elrickvm (approved):
I only had one nit pick on a comment but overall seems good.

app/repositories/packages_repository.rb
sadatay: looks like i snipped a line here by mistake ✂️ thanks for catching that

cowboyd (approved):
I still feel very strongly that we need to move validation inside the repository as we continue moving towards the repository pattern. The reason being that business level validations are the concern of the data owner and shouldn't really be addressed at the controller layer which is serialization/deserialization, encoding / decoding.

Also, for purely practical reasons, validating in the controller actions means that our error handling is fragmented. In other words, we could be error handling once in application controller, or in 30 separate actions spread across 10 controllers.

I'll file a ticket to unify error handling in the controller layer (I think it's pretty important) and then we can go ahead and merge this.

cowboyd:
https://issues.folio.org/browse/UIEH-399

PR #147, opened 2018-05-30
UIEH-277: First pass at RAML for providers endpoint

keathleydavidj:
## Purpose
This PR is my first go at documenting mod-kb's `/providers` endpoint requests using RAML 1.0, addressing the ask of [UIEH-277](https://issues.folio.org/browse/UIEH-277). 

## Approach
- Mostly worked through the first RAML tutorial linked below.

#### TODOS and Open Questions
- [ ] Gather and document response shapes/schemas.
- [ ] Pull relevant pieces out into abstractions or partials where it makes sense.

## Learning
[RAML 100 Tutorial](https://raml.org/developers/raml-100-tutorial)
[RAML 200 Tutorial](https://raml.org/developers/raml-200-tutorial)

### DIFF
--- a/ramls/eholdings.raml
+++ b/ramls/eholdings.raml
@@ -8,10 +8,50 @@ documentation:
   - title: mod-kb-ebsco (category)
     content: Implements the eholdings interface using the EBSCO kb as a backend.
 
-
-/eholdings/vendors:
-  displayName: Vendors
-
+traits:
+  indexable:
+    queryParameters: 
+      q:
+        displayName: Query
+        type: string
+        description: Text entered in search field
+        example: Basket Weaving
+        required: true
+      page:
+        displayName: Page
+        type: number
+        description: Page offset from which results are retrieved
+        example: 1
+        minimum: 1
+        required: false
+      count:
+        displayName: Count
+        type: number
+        description: The maximum number of results to return in the response.
+        example: 10
+        minimum: 0
+        maximum: 100
+        required: false
+      sort:
+        displayName: Sort options
+        enum: [name, relevance]
+        description: Option by which results are sorted
+        default: relevance
+        required: false
+  filterable:
+    queryParameters:
+      filter[selected]:
+        displayName: Selection status
+        enum: [true, false, ebsco]
+        description: Filter to narrow down results based on selection status
+        required: false
+      filter[type]:
+        displayName: Content type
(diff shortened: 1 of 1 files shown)


ramls/eholdings.raml
```diff
@@ -7,6 +7,109 @@ documentation:
...
+    queryParameters: 
+      q:
+        displayName: Query
+        type: string
+        description: Text entered in search field
+        example: Basket Weaving
+        required: true
+      page:
+        displayName: Page
+        type: number
```
sduvvuri-ebsco, line 22: If there is a way to provide a min value here for page offset, we should specify that as 1, else RM API will fail

ramls/eholdings.raml
```diff
@@ -7,6 +7,109 @@ documentation:
...
+        description: Page offset from which results are retrieved
+        example: 1
+        required: false
+      count:
+        displayName: Count
+        enum: [10, 20, 30]
```
sduvvuri-ebsco: similarly, for count, I think we can specify any number between 0 and 100

ramls/eholdings.raml
```diff
@@ -7,6 +7,109 @@ documentation:
...
+        displayName: Sort options
+        enum: [name, relevance]
+        description: Option by which results are sorted
+        default: relevance
+        required: false
+  filterable:
+    queryParameters:
+      filter[selected]:
+        displayName: Selection status
+        enum: [true, false, ebsco]
```
sduvvuri-ebsco, line 45: at least in packages, we support [true, false, ebsco, all]; is that different for providers?

sduvvuri-ebsco:
Overall, this is great @keathleydavidj

ramls/eholdings.raml
keathleydavidj: The way I understood this one when I was playing around with it, selecting `all` just causes this query parameter to be omitted from the request.

ramls/eholdings.raml
keathleydavidj: Reviewing the specification docs revealed that there is indeed a way to restrict the available numbers using `minimum` and `maximum`. -->  https://github.com/raml-org/raml-spec/blob/master/versions/raml-10/raml-10.md/#number
I have gone ahead and updated with those changes.

sduvvuri-ebsco (approved):
Looks like we have to update error messages and figure out how to re-use traits for packages. But, I do not want to hold off this PR for that. We shall deal with it in a separate PR. Great job @keathleydavidj

PR #148, opened 2018-05-30
UIEH-377: API documentation for packages

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-377, we are supposed to hand-roll RAML to document the `/packages` endpoint. This is another attempt to hand-roll some RAML and know if I am moving in the right direction.

## Approach
- Following https://raml.org/developers/raml-100-tutorial, hand-roll RAML for GET /packages and GET /packages/{packageId}
- Earlier PR followed https://raml.org/developers/raml-200-tutorial, trying to make RAML reusable but given our API, which is not completely RESTful(assuming that GET and POST have same schema), I am not sure we can take that approach. 
- So, modified to stick to verbose but more suitable 100-tutorial

#### TODOS and Open Questions
- [ ] Haven't created/used any traits yet
- [ ] Do we need to do something special for `application/vnd.api+json`?

## Learning
https://raml.org/developers/raml-100-tutorial
https://raml.org/developers/raml-200-tutorial

## Screenshots

### DIFF
--- a/ramls/eholdings.raml
+++ b/ramls/eholdings.raml
@@ -1,7 +1,8 @@
-#%RAML 0.8
-title: "mod-kb-ebsco"
+#%RAML 1.0
+title: mod-kb-ebsco
 baseUri: https://github.com/folio-org/mod-kb-ebsco
 version: v1
+mediaType: "application/vnd.api+json"
 
 documentation:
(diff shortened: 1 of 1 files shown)


keathleydavidj (approved):
This is awesome work. Good job!

PR #150, opened 2018-05-31
Part of UIEH-377: Update error responses for packages endpoint

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-377, we are supposed to create documentation to support `/packages` endpoint which has been addressed in this PR https://github.com/folio-org/mod-kb-ebsco/pull/148 
This PR updates error responses after another PR for fixing the error messages has been merged.

## Approach
- Look at new error responses generated by mod-kb-ebsco latest master and update RAML accordingly.

### DIFF
--- a/ramls/eholdings.raml
+++ b/ramls/eholdings.raml
@@ -226,7 +226,7 @@ documentation:
                 {
                   "errors": [
                     {
-                      "title": ""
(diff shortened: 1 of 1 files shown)


keathleydavidj (approved):
Very simple changes here. Looks great!

PR #151, opened 2018-06-01
Fix few issues pointed by raml-cop and add headers info

sduvvuri-ebsco:
## Purpose
Per @dcrossleyau , there is a way to run raml-cop in our local using `./scripts/lint-raml-cop.sh`. 
@dcrossleyau also mentioned that raml-cop reported issues on our master branch. Wonder why it did not fail the build though. Anyways, this PR is an attempt to fix all issues reported by raml-cop.

## Approach
- Run the script, go through each issue in eholdings.raml and fix it.

## Open Questions
[ ] - Is there a way to make `headers:` code reusable since they do not change per request or method?

## Screenshots
screen shot 2018-06-01 at 2 11 51 pm

### DIFF
--- a/ramls/eholdings.raml
+++ b/ramls/eholdings.raml
@@ -34,7 +34,8 @@ traits:
         required: false
       sort:
         displayName: Sort options
-        enum: [name, relevance]
+        type: string
+        enum: ["name", "relevance"]
         description: Option by which results are sorted
         default: relevance
         required: false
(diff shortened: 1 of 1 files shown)


dcrossleyau:
Please note that the script is not hooked in to CI at all yet. At the moment it is just to assist developers, so run it manually as you did.

Hooray, good work -- it is now happy.

sduvvuri-ebsco:
Thanks for keeping us posted about it @dcrossleyau !

keathleydavidj (approved):
This all seems reasonable to me. I did not realize all string values/examples should go in double-quotes, though that kinda seems obvious now.

PR #152, opened 2018-06-05
UIEH-379: RAML for resources endpoint

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-379, we are supposed to hand-roll RAML for `/eholdings/resources` endpoint, this PR solves that purpose.

## Approach
- Investigate various operations permitted and various payloads supported for different HTTP methods for the `/resources` endpoint.
- Code it using RAML.
- Run raml-cop and raml2html locally to ensure that everything looks good.
- Refactor certain object params into `types` section for reusability.
- Gather examples using Advanced Rest Client.

### DIFF
--- a/ramls/eholdings.raml
+++ b/ramls/eholdings.raml
@@ -54,7 +54,45 @@ traits:
         description: Filter to narrow  down results based on content type
         default: all
         required: false
-  
+
+types:
+  customCoverage: # normal object type declaration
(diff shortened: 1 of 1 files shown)


keathleydavidj (approved):
This looks good to me. I like the object types that you define at the top of the file.

PR #153, opened 2018-06-05
UIEH-424: Unit tests for DELETE in resources endpoint

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-424, we have to write unit tests for `DELETE /eholdings/resources` endpoint, which we missed earlier.

## Approach
- Write unit tests for the same.

### DIFF
--- a/spec/requests/resources_spec.rb
+++ b/spec/requests/resources_spec.rb
@@ -766,4 +766,52 @@ RSpec.describe 'Resources', type: :request do
       end
     end
   end
+
+  describe 'deleting a resource' do
+    describe 'delete a custom resource associated with custom package successfully' do
+      before do
+        VCR.use_cassette('delete-custom-resource-custom-package') do
+          delete '/eholdings/resources/123355-2845510-17059786',
+                 headers: okapi_headers
+        end
+      end
+
+      it 'gets a 204 No Content response' do
+        expect(response).to have_http_status(204)
+      end
+    end
+
+    describe 'trying to delete a deleted resource results in error' do
+      before do
+        VCR.use_cassette('delete-deleted-custom-resource') do
+          delete '/eholdings/resources/123355-2843714-17059786',
+                 headers: okapi_headers
+        end
+      end
+
+      let!(:json) { Map JSON.parse response.body }
+
+      it 'gets a not found response' do
+        expect(response).to have_http_status(404)
+        expect(json.errors.first.title).to eq 'Title not found'
+      end
+    end
+
+    describe 'trying to delete a resource associated with managed package' do
+      before do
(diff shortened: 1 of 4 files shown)


spec/requests/resources_spec.rb
```diff
@@ -766,4 +766,52 @@
...
+  describe 'deleting a resource' do
+    describe 'delete a custom resource associated with custom package successfully' do
+      before do
+        VCR.use_cassette('delete-custom-resource-custom-package') do
+          delete '/eholdings/resources/123355-2845510-17059786',
+                 headers: okapi_headers
+        end
+      end
+
+      it 'gets a successful response' do
```
sadatay: i would be specific here about the `204 No Content` rather than `success` which hits for anything in the 200 range.  particularly because this is one of only a few places where we return a 204 right?

spec/requests/resources_spec.rb
```diff
@@ -766,4 +766,52 @@
       end
     end
   end
+
+  describe 'deleting a resource' do
+    describe 'delete a custom resource associated with custom package successfully' do
+      before do
+        VCR.use_cassette('delete-custom-resource-custom-package') do
```
sadatay, line 773: i've been trying to figure out a good solution for this lately while hacking on rspec/vcr on the side.

these tests are all sort of dependent on 
a) the state of _real_ records in the system which are subject to mutation by the tests themselves and by devs messing around in the env - specifically the state of those records at the time of recording.  so you may re-record a test and, in the case of this test, you'd get a 404 instead of a 204 since you already deleted that record the first time you recorded the cassette.  

b) the order in which they are run.  i was always taught that an rspec test should be able to run in any order wtihin its context.  in fact you can turn randomized order on in the sepc_helper to test this theory if i'm remembering correctly... 

anyway, a similar block is at the end of packages_spec too and it's been bugging me.  re-recording the whole suite would error without fail in the last few tests, which was full of these dependent tests.  for most of them earlier on where we're just doing GETs we would have to adjust a few numbers to match newer ones or manually select something only to have the test deselect it while VCR "tape is rolling" so to speak.
for updates, we need to be careful.  

we're not cleaning up any artifacts left in our slice of RMAPI, which is awfully bas citizenship and dangerous.  and we're not setting up any kind of initial state before testing against explicit state values.  most people would claim both of those things as successes.  but this "request test" paradigm isn't your average BDD and i think if we're going to do this against the real API for the recordings we'd better throw some hooks around certain (most?) tests that run _only_ for a re-record.

obviously out of scope here, but figured it was worth noting in case you had any ideas

sadatay (approved):

PR #155, opened 2018-06-05
Return pluralized data type

elrickvm:
## Purpose
Return data type that is pluralized as this contains a collection of things and to keep the inflection consistent.

## Approach
 - Change `ProxyType` => `ProxyTypes`

### DIFF
--- a/app/serializable/serializable_proxy_type.rb
+++ b/app/serializable/serializable_proxy_type.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class SerializableProxyType < SerializableJSONAPIResource
-  type 'proxyType'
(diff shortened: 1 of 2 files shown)


config/routes.rb
```diff
@@ -36,7 +36,7 @@
 
     resources :proxy_types,
               path: 'proxy-types',
-              only: [:index]
+              only: %i[index]
```
sadatay: we should disable this lame rubocop rule.  in what world is this stylistically preferable to the line before it?

sadatay (approved):
lgtm 🚢

config/routes.rb
elrickvm: If it is a single it could be either

PR #157, opened 2018-06-06
Cleanup headers RAML

sduvvuri-ebsco:
## Purpose
I forgot the fact that mod-kb-ebsco is only accessible through Okapi in Folio. So, by the time a request reaches the module, it should have passed mod-auth and mod-permissions, which would have already checked if x-okapi-token, x-okapi-url etc. are present in headers. We need not check for that again.
Also, content-type needs to be application/vnd.api+json only for PUT and POST requests, GETs and DELETEs don't need them. Cleanup headers per aforementioned. 

## Approach
- Cleanup unnecessary header information in RAML.
- Run raml-cop to ensure that its valid RAML.
- Generate HTML locally using raml2html and ensure everything looks as expected.

### DIFF
--- a/ramls/eholdings.raml
+++ b/ramls/eholdings.raml
@@ -97,15 +97,6 @@ types:
   displayName: Packages
   get:
     description: Retrieve a collection of packages
-    headers:
-      X-OKAPI-URL:
-        example: https://sampleurl.io
-      X-OKAPI-TENANT:
-        example: sample tenant
-      X-OKAPI-TOKEN:
-        example: aabbccddee
-      Content-Type: 
-        example: application/vnd.api+json
     queryParameters: 
       q:
         displayName: Search query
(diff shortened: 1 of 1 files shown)


ramls/eholdings.raml
```diff
@@ -228,12 +219,6 @@ types:
   post:
     description: Create a custom package
     headers:
-      X-OKAPI-URL:
-        example: https://sampleurl.io
-      X-OKAPI-TENANT:
-        example: sample tenant
-      X-OKAPI-TOKEN:
-        example: aabbccddee
       Content-Type:
```
keathleydavidj, line 222: I notice that on some of these you leave the `Content-Type:` in where others you remove the whole block. What is the difference?
EDIT: Didn't read the pr description, it is clear why this is.

keathleydavidj (approved):

ramls/eholdings.raml
sduvvuri-ebsco: awesome 👍

PR #159, opened 2018-06-07
UIEH-391: Add RAML examples for requests and code cleanup

sduvvuri-ebsco:
## Purpose
Per UIEH-391, we wanted to add more examples for request payloads in PUT and POST requests so that its easier to come up with payloads for consumers, since we use JSON API spec.

## Approach
- Add examples for request payloads (Note: I touched only `/packages` and `/resources` endpoints so far because I know that there is WIP for the `/providers` endpoint, did not want to step on it and create conflicts).
- Cleanup existing examples by moving them into separate files and including them in existing RAML
- Ensure that its all valid by running raml-cop script

### DIFF
--- a/ramls/eholdings.raml
+++ b/ramls/eholdings.raml
@@ -153,69 +153,12 @@ types:
         description: OK
         body:
           application/vnd.api+json:
-            example: |
-              {
-                "data": [
-                  {
-                    "id": "0-1117849",
-                    "type": "packages",
-                    "attributes": {
-                      "contentType": "Unknown",
-                      "customCoverage": {
-                        "beginCoverage": "",
-                        "endCoverage": ""
-                      },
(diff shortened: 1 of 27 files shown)


sadatay:
sooo happy to have those broken out.  makes `eholdings.raml` actually navigable.  ✨ awesome work and thank you very much for it! ✨ 

what are your thoughts on splitting that main file out based on resource type also?  i was thinking last night maybe it could help compartmentalize some of the places where the service isn't completely RESTful?  could be off base, i dunno.

sduvvuri-ebsco:
I am pretty sure it could be done, just need to figure out how. Will spend some time on it today but it will be in another PR. would like to get this in as we could use this as a pattern for examples going forward.

sadatay (approved):

PR #160, opened 2018-06-07
Fix raml-cop issues and modify examples for request payloads

sduvvuri-ebsco:
## Purpose
Noticed that raml-cop was showing issues at the latest master for one of the `providers` examples.
Also, noticed that if we do not have a "type" attribute in the request payload, then the request to mod-kb fails. 

## Approach
- Fix raml-cop issue
- Add "type" attribute to all request payloads and schema
- Add request example for `providers` and split out examples to follow the new pattern we established.

### DIFF
--- a/ramls/eholdings.raml
+++ b/ramls/eholdings.raml
@@ -153,12 +153,12 @@ types:
         description: OK
         body:
           application/vnd.api+json:
-            example: !include /examples/packages/packages_get_200_response.json        
+            example: !include examples/packages/packages_get_200_response.json        
       400:
         description: Bad Request
         body:
           application/vnd.api+json:
-            example: !include /examples/packages/packages_get_400_response.json
+            example: !include examples/packages/packages_get_400_response.json
   post:
     description: Create a custom package
     headers:
@@ -172,6 +172,7 @@ types:
             type: object
             required: true
             properties:
+              type: string
               attributes:
                 description: Needed because of JSON API
                 type: object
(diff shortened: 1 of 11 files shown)


dcrossleyau:
I noticed that you have some "example: !include ..." entries that have a leading slash on the include path and some that do not.

Do you know what is the best practice (i do not hence the question)?

All of our other RAMLs have no leading slash.

sduvvuri-ebsco:
The paths with leading '/' are treated as absolute paths and the ones without are treated as relative paths. It makes more sense to not have the leading '/' in our case since we want the relative path with respect to the current directory. However, using either did not make a difference in our case since both examples and eholdings.raml are at the same level, I think. I will change them all to be consistent and relative.

sduvvuri-ebsco:
Also, one other question, would it be possible to integrate raml-cop into ci/cd pipeline? Is there anything that is stopping us from doing that? The reason I am asking is because it could be pretty easy for a developer to forget running the script locally before merging code in. Please let me know your thoughts, thanks!

dcrossleyau:
We are working towards that possibility. The first step is to develop a tool that developers can also run locally. It also needs to operate on all repositories, where some have different directory layouts, and some need to exclude certain file/directories. Not yet ready, but see FOLIO-1264.

sduvvuri-ebsco:
Thanks @dcrossleyau

sadatay:
Looks groovy.  We chatted offline about the file trailing newline thing, but otherwise LGTM 👍

sadatay (approved):

keathleydavidj (approved):
This looks great. You've cleaned up the providers section tremendously by pulling out those examples. A++

PR #161, opened 2018-06-11
UIEH-421: Validate resource id for GET, PUT and DELETE

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-421, we were not providing expected error responses when user makes a request with invalid resource id. 

## Approach
- Validate resource id to consist of provider_id, package_id and title_id
- Also, validate that provided values can be converted to integers
- Add associated unit tests

#### TODOS and Open Questions
- [ ] We might want to do something similar for packages, providers and titles endpoints as well. Will discuss with Khalilah and add more bugs in backlog. Added UIEH-425, UIEH-426 and UIEH-427 in backlog.

## Screenshots
resource_id_validation

### DIFF
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -79,7 +79,15 @@ class ResourcesController < ApplicationController
   private
 
   def set_resource
-    @resource = resources.find resource_id
+    resource_id_validation =
+      Validation::ResourceID.new(resource_id)
+
+    if resource_id_validation.valid?
+      @resource = resources.find resource_id
+    else
+      render jsonapi_errors: resource_id_validation.errors,
+             status: :bad_request
+    end
   end
 
   def resource_id
--- a/app/controllers/validation/resource_id.rb
+++ b/app/controllers/validation/resource_id.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+module Validation
+  class ResourceID
+    include ActiveModel::Validations
+
+    attr_accessor :vendor_id, :package_id, :title_id
+
+    validates :vendor_id, presence: true
+    validates :package_id, presence: true
+    validates :title_id, presence: true
+
+    validate :vendor_id_int?
+    validate :package_id_int?
+    validate :title_id_int?
+
+    def vendor_id_int?
+      errors.add(:vendor_id, ':Invalid vendor id') unless
+        vendor_id.to_i != 0
+    end
+
+    def package_id_int?
+      errors.add(:package_id, ':Invalid package id') unless
+        package_id.to_i != 0
+    end
+
+    def title_id_int?
+      errors.add(:title_id, ':Invalid title id') unless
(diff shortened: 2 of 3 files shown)


elrickvm:
Would this handle the failure of when a vendor, package, or title `looks` valid and sends the request to rmapi that then fails.?

elrickvm (commented):
Would this handle the failure of when a vendor, package, or title looks valid and sends the request to rmapi that then fails.? For example something like this... https://folio.frontside.io/eholdings/resources/565-4059099555-427051

I assume it is fine as it will do what it was always doing and 404 but i am not sure if that was error case that we were trying to handle as well.

sduvvuri-ebsco:
very good question @elrickvm , I assumed that RM API would give an error message for "Package not found" and tested mod-kb with your example on this PR, it gives a 404 not found and a HTML message, then I tested RM API and it gives the same, so mod-kb is bubbling up whatever RM API is giving as an error response but maybe we should file a bug with RM API team that they are providing an HTML error response?
Response we get from RM API: 
```<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
<html xmlns="http://www.w3.org/1999/xhtml">
<head>
<meta http-equiv="Content-Type" content="text/html; charset=iso-8859-1"/>
<title>404 - File or directory not found.</title>
<style type="text/css">
<!--
body{margin:0;font-size:.7em;font-family:Verdana, Arial, Helvetica, sans-serif;background:#EEEEEE;}
fieldset{padding:0 15px 10px 15px;} 
h1{font-size:2.4em;margin:0;color:#FFF;}
h2{font-size:1.7em;margin:0;color:#CC0000;} 
h3{font-size:1.2em;margin:10px 0 0 0;color:#000000;} 
#header{width:96%;margin:0 0 0 0;padding:6px 2% 6px 2%;font-family:"trebuchet MS", Verdana, sans-serif;color:#FFF;
background-color:#555555;}
#content{margin:0 0 0 2%;position:relative;}
.content-container{background:#FFF;width:96%;margin-top:8px;padding:10px;position:relative;}
-->
</style>
</head>
<body>
<div id="header"><h1>Server Error</h1></div>
<div id="content">
 <div class="content-container"><fieldset>
  <h2>404 - File or directory not found.</h2>
  <h3>The resource you are looking for might have been removed, had its name changed, or is temporarily unavailable.</h3>
 </fieldset></div>
</div>
</body>
</html>
```

elrickvm:
Cool. I know with the new pattern we have more control over request failures from RMAPI. But i think this works for this for sure.

elrickvm (approved):

sduvvuri-ebsco:
@elrickvm : I sent an email to Mohan from RM API team to see if he can file a bug or if that is intentional. Thanks for your feedback!

PR #163, opened 2018-06-12
UIEH-369: Handle error responses in packages consistently as titles

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-369, there is a discrepancy in error response handling in mod-kb-ebsco when we are trying to create custom packages versus. custom titles. This PR tries to give consistent error responses so that it does not mess up UI toast messages.

## Approach
- Figure out if RM API is giving error message in a consistent format and in this case, yes, RM API gives error responses in consistent format while creating custom titles and custom packages.
- Figured out that the discrepancy is indeed in mod-kb, probably introduced with the new repository pattern followed.
- Figured that RM API is giving error responses in JSON format where as there are quite a few error responses created in mod-kb-ebsco which are not in JSON and are plain strings for validations done within our code.
- Due to the above difference, added exception handling around the JSON.parse block and modified the way we support error response handling in `catch_repository_errors`.
- Ensured that all unit tests are passing and we are getting error response in expected format while trying to create a custom package whose name already is used.

#### TODOS and Open Questions
- [ ] Probably, there is a cleaner way to do this in the RequestError class at a higher level, or we could modify all of our internally generated error to be JSON also? ideas welcome. Probably, when we change other endpoints to use this repository pattern, then this difference in error handling needs to be converged.

## Screenshots
error_handling_consistency

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -43,8 +43,16 @@ class ApplicationController < ActionController::API
   def catch_repository_errors
     yield
   rescue RmapiRepository::RequestError => e
-    render jsonapi_errors: [title: e.message],
-           status: e.status
+    begin
+      # If the error is from RM API, its in JSON
+      json = JSON.parse(e.message)
+      render jsonapi_errors: get_json_error_hash(json),
+             status: e.status
+    rescue JSON::ParserError
+      # If the error is from within our code, its a plain string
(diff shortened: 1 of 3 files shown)


carolegodfrey (approved):
LGTM - Nice job Soba - especially liked confirming things with telepresence!

PR #164, opened 2018-06-13
UIEH-392 Validate Package URL Query Parameters

elrickvm:
## Purpose
When making a search request a Client has the ability to send URL Query Parameters along with the request to filter down the search result or data set that is returned. This request and Query Parameters are then translated and sent off to RMAPI who then responds with the results or data that the client requests. As long as the Query Parameters values are acceptable with RMAPI then all is fine. However, when the values that sent under Query Parameters those would pass through and be sent to RMAPI which would sometimes cause RMAPI to error out, or respond with a 200 but actually not accepting the invalid Query Parameters. We are making a spot gap before those invalid Query Parameters are requested to RMAPI.

### JIRA
https://issues.folio.org/browse/UIEH-392

## Approach
Validate Each accepted Query Parameter

## Screenshots
Example of passing invalid Query Parameters
screen shot 2018-06-13 at 7 16 36 pm


app/controllers/validation/package_query_parameters.rb
```diff
@@ -0,0 +1,41 @@
...
+    def initialize(params = {})
+      filters = params.fetch(:filter, nil)
+      @sortFilter = params[:sort]
+      @selectedFilter = hash?(filters) ? get_filter(filters, :selected) : filters
+      @contentTypeFilter = hash?(filters) ? get_filter(filters, :type) : filters
+    end
+
+    private
+
+    def get_filter(hash, key)
```
sadatay: thiiis might be overkill. i'm all for single responsibility principle but hash already has the responsibility of doing hash lookups so why not just hit it head-on whence this method is invoked.


lil_tree

sadatay:
so most APIs would just ignore invalid query params - i'm not sure how this made it to implementation without someone piping up about that but i'm sorry for missing it earlier.  if we do want to implement it for some reason, this is a reasonable way to do it and is tested from all angles 👍.  i just think it looks like a bike shed no matter how you approach it

sadatay (approved):

sduvvuri-ebsco (approved):
Looks pretty good Elrick!

PR #165, opened 2018-06-13
UIEH-378: RAML Descriptors for Title Resources

sadatay:
We are using RAML to express our API contract in a commonly consumed
format that is spoken in one way or another by many modern web
development ensembles.  We can then publish this as a both a source of
human-readable documentation for the non `ui-eholdings` API consumer
and a schema that can be consumed by tooling to create an interface
betwen modules or even to draft an outline of a module - anything
available to RAML is now available as an option we can offer to users.

This work covers the endpoints related to Title resources, based at `/eholdings/titles`.


elrickvm (approved):

ramls/examples/titles/titles_titleId_get_404_response.json
```diff
@@ -0,0 +1,5 @@
+{
+  "jsonapi": {
```
sduvvuri-ebsco: Don't we get any error response here? Like "Title Not Found" or similar?

sduvvuri-ebsco (commented):
Similarly, the example of response for titles_post_422_response.json might be incomplete. 
Also, wanted to just double check that we are running raml-cop script locally `./scripts/lint-raml-cop.sh` before merging it in. Overall, looks pretty good.

ramls/examples/titles/titles_titleId_get_404_response.json
sadatay: evidently not!  i have that marked down as minor bug to add today.  

screen shot 2018-06-14 at 1 05 14 pm

sadatay:
Fixed linting issues after getting bash 4 running.

While there is no 422 response represented here, it is because i could not find a path of code execution that would actually trigger it.  The two pathways one might use are:

1. Providing a payload that fails validation, but I was not able to get this to occur.  With valid json the endpoint seemingly accepted anything - it appears that an outdated validation file is being applied but why it passes I do not know.  Will be filing this behavior as a bug tonight.
2. Trigger a `JSON::ParseError` - any malformed JSON we send gets caught in the application controller, which hacks up a 404 before we get to any controller code.  

So functionally there are no 422s for this endpoint at this moment.

PR #166, opened 2018-06-14
Return a 400 Error response for invalid query parameters

elrickvm:
## Purpose
If a Client pass values in Query Parameters that are not valid or acceptable value return a 400 Error response with the Error Message.


sduvvuri-ebsco (approved):
Awesome, thanks @elrickvm!

sadatay (approved):
💯 x 4

PR #167, opened 2018-06-27
UIEH-435 Validate query params for titles endpoint

elrickvm:
## Purpose
When making a search request a Client has the ability to send URL Query Parameters along with the request to filter down the search result or data set that is returned. This request and Query Parameters are then translated and sent off to RMAPI who then responds with the results or data that the client requests. As long as the Query Parameters values are acceptable with RMAPI then all is fine. However, when the values that sent under Query Parameters those would pass through and be sent to RMAPI which would sometimes cause RMAPI to error out, or respond with a 200 but actually not accepting the invalid Query Parameters. We are making a spot gap before those invalid Query Parameters are requested to RMAPI.

## Approach
- Validate URL Query Parameters for title endpoint and return a 400 request error if the Query Params fails the validation

## Screenshots
Example of sending over Query Params that are allowed.
screen shot 2018-06-27 at 4 01 00 pm


sduvvuri-ebsco (approved):
Great work! Thanks for doing this Elrick!

elrickvm:
Thank You! @sduvvuri-ebsco and you're welcome.

PR #168, opened 2018-06-27
Add raml-cop to Travis build

cherewaty:
screen shot 2018-06-27 at 3 16 23 pm

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,9 @@ rvm:
 
 install:
   - bundle install
+  - yarn global add raml-cop
 
 script:
   - rubocop
+  - ./scripts/lint-raml-cop.sh
   - COVERAGE=true rake spec


.travis.yml
```diff
@@ -8,7 +8,9 @@ rvm:
 
 install:
   - bundle install
+  - yarn global add raml-cop
```
sduvvuri-ebsco, line 11: This might be a really dumb question but can we add raml-cop using yarn? I have seen it to be part of npm?

sduvvuri-ebsco (approved):

.travis.yml
sduvvuri-ebsco: never mind, just saw your screenshot

cherewaty: Any `npm` package can be installed with `yarn`. `yarn` is really just a drop in replacement for the `npm` CLI - they talk to the same registry through different proxies.

sduvvuri-ebsco: Ah, good to know. Did not know that, thanks @cherewaty !

PR #170, opened 2018-06-28
[UIEH-397] Refactor Title POST tests to support full payload

SamKeathley:
## Purpose
Refactor tests so that they reflect against a real-world scenario for reliability. 
Resolves [UIEH-397](https://issues.folio.org/browse/UIEH-397)

## Approach
The major change is that the full request payload one should expect when creating a title is a hash that is referenced throughout the entire test file and ran with each individual test. Before, we were asserting only relevant items on a test-by-test basis or "hard-coding" in solutions for the sake of a test.


cowboyd (approved):
Fantastic work @SamKeathley.

PR #171, opened 2018-07-09
FOLIO-1179 fix dev.f.o URLs

dcrossleyau:


dcrossleyau:
Fixes old dev.folio.org URLs, and to be https.

Links to the general "FOLIO" issue tracker project. There does not seem to be a dedicated Jira Project, as is for other modules. Is there is, then please link directly to it.

dcrossleyau:
Ping. Please someone approve.

cherewaty (approved):

PR #173, opened 2018-07-13
UIEH-429: Add support for advanced title search in mod-kb-ebsco

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-429, RM API now supports advanced search while getting titles. Add support for the same.

## Approach
- Whenever mod-kb-ebsco makes a request to RM API to get a collection of titles, always set advancedsearch=true and pass it through in the request.
- Record new VCRs and modify tests according to the new results.

## Screenshots
advanced_search


carolegodfrey (approved):
LGTM Sobha!

PR #177, opened 2018-07-23
UIEH-484: Update resources POST request to include url as parameter

carolegodfrey:
## Purpose
https://issues.folio.org/browse/UIEH-484

The Resources POST request creates a new resource (essentially associating a title with a package)
Per documentation and observed UI eHoldings requests, attributes passed with this request include (packageId, titleId and url)

https://s3.amazonaws.com/foliodocs/api/mod-kb-ebsco/eholdings.html#eholdings_resources_post 

When testing mod-kb-ebsco API POST request, we found that a url passed into the request was being ignored. 

Additionally, the url passed to the request should be verified as valid (validation is minimal, confirming url starts with either https:// or http://)

A sample payload is as follows:
` {"data":{"type":"resources","attributes":{"url": "http://test.io", "packageId":"123355-3111169","titleId":"13493928"}}}`

## Approach
Added url to allowed parameters for POST request
Added validation of url (starts with either http:// or https://)
Added tests (with and without a url and with an invalid url)

#### TODOS and Open Questions
- [ ] Resources PUT request requires validation of the URL -- this is a separate ticket (https://issues.folio.org/browse/UIEH-495)
- [x] Resolve rubocop -- app/controllers/resources_controller.rb:11:3: C: Metrics/AbcSize: Assignment Branch Condition size for create is too high. [31.34/30]

### DIFF
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -8,6 +8,8 @@ class ResourcesController < ApplicationController
   deserializable_resource :resource,
                           only: %i[create update],
                           class: DeserializableResource
+
+  # rubocop:disable Metrics/AbcSize
   def create
     # A resource represents the relationship between a package
     # and title.  So in this `create` method we're really just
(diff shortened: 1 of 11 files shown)


sduvvuri-ebsco (approved):
Looks good @carolegodfrey, build has not succeeded though. Also, curious about rubocop issue. Using our judgement, we either break up that code into smaller chunks or disable rubocop warning for that piece of code if we are sure we want to keep it together.

PR #179, opened 2018-07-26
UIEH-500: Remove all code related to the old unused 'vendors' endpoint

keathleydavidj:
Resolves: [UIEH-500](https://issues.folio.org/browse/UIEH-500)

## Purpose
Some time ago mod-kb-ebsco's `/vendors` endpoint was phased out in favor of a `/providers` endpoint for better semantics. This PR is meant to remove all of the now unused code left over from that name change, reducing old cruft and module size.


spec/requests/packages_spec.rb
```diff
@@ -460,33 +460,6 @@
     end
   end
 
-  describe 'getting a package with included vendor' do
```
keathleydavidj, line 463: These were the only places I had any uncertainty about. I'm not sure if these tests needed to be removed or if they should have been changed to test for `?include=provider` instead.

spec/requests/resources_spec.rb
```diff
@@ -133,33 +133,6 @@
     end
   end
 
-  describe 'getting a resource with included vendor' do
```
keathleydavidj, line 136: Same as above.

sduvvuri-ebsco: @keathleydavidj : Please check if there are similar tests for provider. For Example: `getting a package with included provider` and `getting a resource with included provider`. If there are, then these can safely be deleted. Else, please adjust them to look for `provider` instead. Hope this helps. Other than verifying the open questions you have on this PR, the changes look good to me!

carolegodfrey (approved):
LGTM!

PR #180, opened 2018-07-31
UIEH-483 - Packages filter[custom] update documentation/validation of true as only value

carolegodfrey:
## Purpose
https://issues.folio.org/browse/UIEH-483

While writing API tests for Packages endpoint, the following was discovered:
Per documentation here: https://s3.amazonaws.com/foliodocs/api/mod-kb-ebsco/eholdings.html#eholdings_packages_get; filter[custom] can have 2 values, true or false but passing in either is not making a difference in the response. 

The custom filter is used to populate the dropdown for getting the list of custom packages (when associating a title to a package)  -- we use the filter[custom]=true in this case from eHoldings. The code in mod-kb-ebsco only supports getting custom packages (based on this filter).  There is not a use case to for filter[custom]= false to get a list of packages that are not custom. 

This PR updates documentation and validation for filter[custom] so that it is not ambiguous when trying to use it.

## Approach
- Update the documentation to indicate filter[custom]=true is the only allowed value.
- Update mod-kb-ebsco to validate that the filter passed has value of true.

### DIFF
--- a/app/controllers/validation/package_query_parameters.rb
+++ b/app/controllers/validation/package_query_parameters.rb
@@ -7,7 +7,7 @@ module Validation
     include ActiveModel::Validations
 
     attr_accessor :sortFilter, :selectedFilter,
-                  :contentTypeFilter
+                  :contentTypeFilter, :customFilter
 
     validates :sortFilter, inclusion: { in: %w[name relevance],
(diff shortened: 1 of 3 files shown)


sduvvuri-ebsco (approved):
Looks good @carolegodfrey!

PR #181, opened 2018-07-31
UIEH-463: Update RAML for endpoints with collections

keathleydavidj:
Resolves: [UIEH-463](https://issues.folio.org/browse/UIEH-463)

## Purpose
  This PR is meant to address some missed points in the mod-kb-ebsco RAML documentation.

## Approach
```
GET /eholdings/packages
GET /eholdings/providers
GET /eholdings/titles
```
- Fixes `q` query param being marked as required when it isn't
- Adds the default value for the `count` query param

`GET /eholdings/providers/{provider_id}`
- Clarify that "packages" is the only recognized value for the `include` query param

### DIFF
--- a/ramls/eholdings.raml
+++ b/ramls/eholdings.raml
@@ -16,7 +16,7 @@ traits:
         type: string
         description: Text entered in search field
         example: Basket Weaving
-        required: true
+        required: false
       page:
         displayName: Page
         type: number
@@ -31,6 +31,7 @@ traits:
         example: 10
         minimum: 0
         maximum: 100
+        default: 25
         required: false
       sort:
         displayName: Sort options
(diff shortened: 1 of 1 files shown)


ramls/eholdings.raml
```diff
@@ -16,7 +16,7 @@ traits:
         type: string
         description: Text entered in search field
         example: Basket Weaving
-        required: true
+        required: false
```
carolegodfrey: This looks good --One small item -  I think we want to apply this to packages and titles endpoints as well

carolegodfrey, line 19: Looks good -- just one small comment --think this needs to be applied to indexable trait -- since it should apply to packages/providers and titles (both required and default)

carolegodfrey (approved):
LGTM @keathleydavidj

PR #182, opened 2018-07-31
UIEH-491: Bug fix for putting a custom/managed resource

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-491, there is an issue in mod-kb updating managed titles, where we provide fields that can be updated only for custom resources and the response is a 200 showing the updated values as passed in the request. However, when we make a GET request on the same `titleId`, we get the correct response. So, the response after PUT request was totally misleading. Fix that in this PR.

## Approach
- Identify based on https://s3.amazonaws.com/foliodocs/api/mod-kb-ebsco/eholdings.html#eholdings_resources__resourceid__put what fields can be updated only for custom vs. managed resources.
- Add validations in resource updates to ensure that fields that are not expected are indeed not present in the request body.
- Modify the update method in the resource model to update certain fields in RM API only if they are valid for the custom resource, else skip that step and only deal with updates to managed resource fields.
- Modify associated unit tests accordingly.

## Screenshots
put_resources

### DIFF
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -58,8 +58,9 @@ class ResourcesController < ApplicationController
   end
 
   def update
+    is_title_custom = title_custom?
     resource_validation =
-      Validation::ResourceUpdateParameters.new(update_params)
+      Validation::ResourceUpdateParameters.new(is_title_custom, update_params)
 
     if resource_validation.valid?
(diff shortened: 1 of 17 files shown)


carolegodfrey (approved):
Changes look good to me @sduvvuri-ebsco

PR #183, opened 2018-08-01
UIEH-495: Fix invalid url in PUT request for resources

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-495, mod-kb allows invalid url to be passed in while updating a resource to RM API which is not expected behavior. Add validation to avoid that.

## Approach
- Tried to figure out if this validation needs to be done in the creation of a custom resource as well and saw that its being validated there.
- Kept the approach the same for both creation and update.
- Wrote an associated unit test

## Screenshots
validate_url


carolegodfrey (approved):
Looks Good!

PR #184, opened 2018-08-01
UIEH-470: Validate filters within packages associated with a provider

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-470, we were allowing invalid values for filters in query params for list of packages associated with a given provider. This PR fixes that issue.

## Approach
- Looking at docs here: https://s3.amazonaws.com/foliodocs/api/mod-kb-ebsco/eholdings.html#eholdings_providers__provider_id__packages_get and validating what filters are allowed in UIEH for search within packages for a given provider, understand what values are allowed for various filters.
- Validate them in the ProvidersController
- Write associated unit tests

## Screenshots
validate_filter_params

### DIFF
--- a/app/controllers/validation/provider_packages_query_parameters.rb
+++ b/app/controllers/validation/provider_packages_query_parameters.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+# rubocop:disable Naming/VariableName
+
+module Validation
+  class ProviderPackagesQueryParameters
+    include ActiveModel::Validations
+
+    attr_accessor :sortFilter, :selectedFilter,
+                  :contentTypeFilter
+
+    validates :sortFilter, inclusion: { in: %w[name relevance],
+                                        message: 'Invalid Query Parameter for sort' }, allow_nil: true
+    validates :selectedFilter, inclusion: { in: %w[true false ebsco all],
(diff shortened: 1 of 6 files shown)


app/controllers/validation/provider_packages_query_parameters.rb
```diff
@@ -0,0 +1,37 @@
...
+    include ActiveModel::Validations
+
+    attr_accessor :sortFilter, :selectedFilter,
+                  :contentTypeFilter
+
+    validates :sortFilter, inclusion: { in: %w[name relevance],
+                                        message: 'Invalid Query Parameter for sort' }, allow_nil: true
+    validates :selectedFilter, inclusion: { in: %w[true false ebsco],
+                                            message: 'Invalid Query Parameter for filter[selected]' }, allow_nil: true
+    validates :contentTypeFilter, inclusion: { in: %w[aggregatedfulltext abstractandindex ebook ejournal print onlinereference unknown],
```
carolegodfrey: This all looks great @sduvvuri-ebsco -- I think we need all to be included in contentType filter as well though?

carolegodfrey: Looks great @Sobha -- just one small think - I think we need to check for all as well

carolegodfrey (approved):

PR #185, opened 2018-08-01
UIEH-427 Ensure both provider and package ids are in url

elrickvm:
## Purpose
In order to `GET`, `DELETE`, `UPDATE` a package, both the provider_id and package_id have to be present in the URL. 

## Approach
 - Esure that two ids are passed in the URL and return an Error if not

#### Open Questions
- [ ] There was discussion around testing that the URL provider and package id's are only numeric and not alphanumeric. Example `/eholdings/packages/123-123` vs `/eholdings/packages/123-123ABC`. In regards for testing for alphanumeric non-presence. I could not find an overwhelming reason to test for that based on what currently happens. If you pass an id that RMAPI does not find whether it is numeric or alphanumeric it will return 404 not found which seems acceptable. 

## Screenshots
screen shot 2018-08-01 at 3 12 51 pm

### DIFF
--- a/app/repositories/packages_repository.rb
+++ b/app/repositories/packages_repository.rb
@@ -3,6 +3,9 @@
 class PackagesRepository < RmapiRepository
   def find!(id)
     vendor_id, package_id = id.split('-')
+
+    fail RequestError.new('Package and provider id are required', 400) unless vendor_id && package_id
+
     status, body = request(:get, "/vendors/#{vendor_id}/packages/#{package_id}")
     Result.new(data: to_package(body), status: status, included: body[:included])
   end
@@ -62,6 +65,9 @@ class PackagesRepository < RmapiRepository
 
   def destroy!(id)
     vendor_id, package_id = id.split('-')
+
+    fail RequestError.new('Package and provider id are required', 400) unless vendor_id && package_id
+
     status, _no_content = request(:put, "/vendors/#{vendor_id}/packages/#{package_id}", json: { isSelected: false })
 
     Result.new(data: {}, status: status)
--- a/spec/requests/packages_spec.rb
+++ b/spec/requests/packages_spec.rb
@@ -257,6 +257,21 @@ RSpec.describe 'Packages', type: :request do
     it 'returns a valid visibility reason' do
       expect(json.data.attributes.visibilityData.reason).to eq 'Set by system'
     end
+
+    describe 'getting a package with invalid package id' do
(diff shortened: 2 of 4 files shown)


spec/requests/packages_spec.rb
```diff
@@ -257,6 +257,21 @@
     it 'returns a valid visibility reason' do
       expect(json.data.attributes.visibilityData.reason).to eq 'Set by system'
     end
+
+    describe 'with an invalid url missing package id' do
```
sduvvuri-ebsco: Really minor comment @elrickvm : can we rephrase this test name to something else -- `with an invalid url missing package id` is a little confusing because we can pass `url` for certain requests in request body -- probably rename it to something like `getting a package with invalid package id`?

app/repositories/packages_repository.rb
```diff
@@ -3,6 +3,9 @@
 class PackagesRepository < RmapiRepository
   def find!(id)
     vendor_id, package_id = id.split('-')
+
+    fail RequestError.new('Package and vendor id are required', 400) unless vendor_id && package_id
```
sduvvuri-ebsco: Can we also change this error message to `Package and provider id are required` because consumer of mod-kb-ebsco would not know what a vendor is...its an RM API term serialized to `provider` in mod-kb-ebsco

app/repositories/packages_repository.rb
```diff
@@ -62,6 +65,9 @@ def update!(id, attrs)
 
   def destroy!(id)
     vendor_id, package_id = id.split('-')
+
+    fail RequestError.new('Package and vendor id are required', 400) unless vendor_id && package_id
```
sduvvuri-ebsco: Same comment as above in replacing vendor id with provider id

spec/requests/packages_spec.rb
```diff
@@ -1187,6 +1202,22 @@
...
+          delete '/eholdings/packages/123355',
+                 headers: okapi_headers
+        end
+      end
+
+      let!(:json_f) { Map JSON.parse response.body }
+
+      it 'get a error response' do
+        expect(response).to have_http_status(400)
+        expect(json_f.errors.first.title).to eql('Package and vendor id are required')
```
sduvvuri-ebsco: We'll have to adjust these tests after we exchange vendor with provider.

sduvvuri-ebsco (approved):
Looks great!

PR #186, opened 2018-08-02
UIEH-466: Honor count param in providers and titles

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-466, we uncovered a bug in mod-kb-ebsco that providers and titles are not honoring the count parameter being passed in query params. Fix that in this PR.

## Approach
- Looking through code, understood that the param is in fact not being picked up in controllers as a result of which, it is not being sent to model and to RM API.
- Add that logic in there and associated unit tests.

## Screenshots
count_param


elrickvm (approved):
Looks good!

PR #188, opened 2018-08-03
Part of UIEH-427: Validate packageId

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-427, we are supposed to add validation for packageId in GET package with packageId requests. Part of the work has been addressed in https://github.com/folio-org/mod-kb-ebsco/pull/185. There was a corner case identified in the error response being returned not abiding by JSON API schema. This PR addresses that.

## Approach
- Add validation for provider and package ids to contain only digits and not alphabets
- Port that change over to validating resource ids
- Add associated unit tests

## Screenshots
packageid

### DIFF
--- a/app/controllers/validation/resource_id.rb
+++ b/app/controllers/validation/resource_id.rb
@@ -16,17 +16,17 @@ module Validation
 
     def vendor_id_int?
       errors.add(:vendor_id, ':Invalid vendor id') unless
-        vendor_id.to_i != 0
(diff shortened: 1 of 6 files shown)


elrickvm (commented):
I would say just add a test case for sending a request like `abv1217sdsd` that is alphanumeric

elrickvm:
Or even one that starts with `1217ababa`..

elrickvm (approved):

PR #189, opened 2018-08-08
UIEH-264: Fix root-proxy endpoint to match RM API behavior

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-264, we are supposed to ensure that we do not pass any unsafe url characters in the PUT request to `/eholdings/root-proxy`. While there is no scope of doing that in our code, on testing this feature in various hosting environments, identified that GET or PUT `/eholdings/root-proxy` are not working in FSE Hosting/Index Data environment while it works perfectly fine in Frontside environment. On further investigation and on confirming with RM API team, sandbox and production RM API handles `/custid/` endpoint differently. Sandbox needs the trailing slash to get the correct response, giving a 403 otherwise where as Production irrespective of the trailing slash gives a 200 response and this is the exact reason why we were not seeing this issue in Frontside deployed code because it points to Production RM API where as FSE and Index Data  point to Sandbox RM API.

## Approach
- Modify the way we send request URLs for root-proxies and custom-labels to RM API such that irrespective of environment, they are still correct.
- Fix associated unit tests.

## Learning
- A trailing slash in a url could mean a lot...lol

## Screenshots
*Before*
root_proxy_before

*After*
root_proxy_after

### DIFF
--- a/app/repositories/rmapi_repository.rb
+++ b/app/repositories/rmapi_repository.rb
@@ -15,9 +15,13 @@ class RmapiRepository
   end
 
   def request(verb, fragment = '', **options)
+    ## Construction of url is different based on whether fragment is present or not because
+    ## RM API in sandbox needs a trailing slash when fragment is not present.
+    ## As of 08/08/2018, it gives a 403 otherwise.
+    url = fragment == '' ? "#{base_url}/" : "#{base_url}#{fragment}"
     response = HTTP.headers(headers).request(
       verb,
-      "#{base_url}#{fragment}",
+      url,
       options
     )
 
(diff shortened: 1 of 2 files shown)


app/repositories/rmapi_repository.rb
```diff
@@ -15,9 +15,13 @@ def initialize(config:)
   end
 
   def request(verb, fragment = '', **options)
+    ## Construction of url is different based on whether fragment is preesnt or not because
```
carolegodfrey: Extremely Minor nit on spelling - preesnt

sduvvuri-ebsco:
Thanks @carolegodfrey , corrected it!

carolegodfrey (approved):
Looks Good Sobha!

PR #190, opened 2018-08-08
UIEH-375: Add package proxy support

keathleydavidj:
## Purpose
To resolve [UIEH-375](https://issues.folio.org/browse/UIEH-375), users should be able to view the current proxy setting as well as change the selected proxy at the package level. This PR takes care of the back end implementation.

## Approach
- [x] Write unit tests
- [x] Add support for (de)serializing the proxy field on the GET and PUT calls for a package.

## Screenshots
TODO

### DIFF
--- a/spec/requests/packages_spec.rb
+++ b/spec/requests/packages_spec.rb
@@ -265,6 +265,7 @@ RSpec.describe 'Packages', type: :request do
       expect(json.data.id).to eq('19-6581')
       expect(json.data.attributes).to include(
         'name',
+        'proxy',
         'contentType',
         'titleCount',
         'selectedCount',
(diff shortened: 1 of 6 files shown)


spec/requests/packages_spec.rb
```diff
@@ -860,6 +866,51 @@
...
+                "isSelected": true,
+                "isHidden": false,
+                "allowEbscoToAddTitles": true,
+                "customCoverage": {
+                  "beginCoverage": '2003-01-01',
+                  "endCoverage": '2003-12-12'
+                },
+                "proxy": {
+                  "id": 'EZProxy',
+                  "inherited": false
```
sduvvuri-ebsco: @keathleydavidj : just a very minor comment: Let's try to update the `proxy -> id` along-with `inherited` to see if the update is happening as expected in the test. We are aware of few issues with RM API `inherited` and it is not a reliable test. If you make a request to `/eholdings/proxy-types`, it should give you the list of allowed proxies and you can pick one from those for the update test. Looks pretty good otherwise 👍

elrickvm:
Awesome work David!

elrickvm (approved):

sduvvuri-ebsco (approved):
Looks good @keathleydavidj!

PR #191, opened 2018-08-09
UIEH-485: Handle invalid content type header

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-485, we are supposed to handle invalid Content-Type headers passed in request without giving a 500 with a stack trace. This PR addresses that issue.

## Approach
- Get the type of request and if it is a PUT or POST, ensure that the expected Content-Type is passed in headers, else give a bad request with appropriate error message.
- Add associated unit test.

#### TODOS and Open Questions
- [x] Application Controller itself does not have any file containing unit tests. So, I added the test to Configurations although it does not belong there. This can happen across requests in any part of code. Wondering if I should move this test to its own file and add tests for rest of the application controller code or is it an overkill? 🤔 

## Screenshots
content_type_header

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 class ApplicationController < ActionController::API
+  before_action :verify_content_type_header
   before_action :verify_okapi_headers
   before_action :set_response_headers
   around_action :catch_exceptions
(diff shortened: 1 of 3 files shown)


carolegodfrey:
I think having a test on one of the endpoints (like configuration) is sufficient. Maybe adding a POST example in addition to the PUT would be helpful since it would cover the 2 noted types of request for which there is an error. Looks great!

sduvvuri-ebsco:
Sounds good, thanks @carolegodfrey for your feedback. I will probably add a test for POST in an example where POST is allowed like custom packages.

carolegodfrey (approved):

PR #192, opened 2018-08-10
UIEH-489: Validate custom coverage in PUT/POST requests of resources/packages

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-489, we are supposed to validate custom coverage in POST/PUT requests for (custom/managed) resources and packages where applicable so that the consumer gets an understandable error message instead of stack trace with bad request. 

## Approach
- Added necessary validation for the dates themselves both in resources and packages
- When json in the request body is invalid, we used to give a 400 with stack trace. Tweaked middleware for all json requests for better error message reporting when there are JSON parse errors, otherwise it behaves as before.
- Add associated unit tests

#### TODOS and Open Questions
- [ ] Unsure how to write a unit test for invalid json request since rubocop complains while linting the request body. Proof that it works is in the recorded gif and has been manually tested. We could probably also test this when we add more / modify our API tests to test the changes in this PR.

## Learning
https://github.com/rails/rails/blob/fc5dd0b85189811062c85520fd70de8389b55aeb/actionpack/lib/action_dispatch/middleware/show_exceptions.rb

## Screenshots
custom_coverage_validate

### DIFF
--- a/app/controllers/validation/custom_package_parameters.rb
+++ b/app/controllers/validation/custom_package_parameters.rb
@@ -6,10 +6,32 @@ module Validation
   class CustomPackageParameters
     include ActiveModel::Validations
 
-    attr_accessor :name, :contentType
+    attr_accessor :name, :contentType, :beginCoverage, :endCoverage
 
     validates :name, presence: true
     validates :contentType, presence: true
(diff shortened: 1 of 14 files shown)


elrickvm (approved):
This looks good.

PR #194, opened 2018-08-14
FOLIO-722 Add missing LICENSE

dcrossleyau:
Add missing LICENSE file.


LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      License. However, in accepting such obligations, You may act only
+      on Your own behalf and on Your sole responsibility, not on behalf
+      of any other Contributor, and only if You agree to indemnify,
+      defend, and hold each Contributor harmless for any liability
+      incurred by, or claims asserted against, such Contributor by reason
+      of your accepting any such warranty or additional liability.
+
+   END OF TERMS AND CONDITIONS
+
+   APPENDIX: How to apply the Apache License to your work.
```
cherewaty, line 178: Is this appendix supposed to be copied here?

cherewaty (commented):
Should we just be following the instructions at http://www.apache.org/licenses/LICENSE-2.0#apply instead of pasting in the contents of that page?

LICENSE
dcrossleyau: No, that is for other developers beyond folio-org/OLF that utilise this licensed work, to apply the license to their other work.

PR #195, opened 2018-08-16
UIEH-525: Fix issue with updating resources

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-525, we are able to see multiple errors when we try to update a managed resource. Reason being that we added validation in mod-kb-ebsco to let user be notified when they update fields that they are not supposed to for managed resources as addressed in https://github.com/folio-org/mod-kb-ebsco/pull/182. However, since this shows up as multiple toasts in ui, and since UI always sends the complete payload it gets from making a GET request, ignore fields that cannot be updated for a managed resource such that multiple errors do not show up and at the same time, response of PUT is not misleading because fields that cannot be updated result in correct values or be null.

## Approach
- Remove the validation added earlier
- Instead, in the controller, pass only fields that can be updated for custom/managed resources
- Adjust the unit test accordingly.

#### TODOS and Open Questions
- [ ] Fix API tests in FSE Hosting to suit this change.

## Screenshots
resources_fix

### DIFF
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -58,9 +58,8 @@ class ResourcesController < ApplicationController
   end
 
   def update
-    is_title_custom = title_custom?
     resource_validation =
-      Validation::ResourceUpdateParameters.new(is_title_custom, update_params)
+      Validation::ResourceUpdateParameters.new(update_params)
 
     if resource_validation.valid?
       @resource.update update_params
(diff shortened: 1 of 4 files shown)


carolegodfrey (approved):
LGTM @sduvvuri-ebsco   -- so instead of rejecting requests with custom fields that don't apply for managed title -- we now only consider applicable fields (custom or managed) and ignore those that don't apply -- should make things easier on front end

sduvvuri-ebsco:
Yep, exactly. Ideally, we should have had those validations in place and UI should send different payloads for custom/managed but that probably warrants a separate discussion and we can bring it up tomorrow in retrospective.

PR #197, opened 2018-08-22
UIEH-177: GET and PUT of package tokens at package level

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-177, we are supposed to implement GET and PUT of package tokens at package level.

## Approach
- Add support for GET and PUT of package tokens at package level
- Uncover an RM API issue with updating package tokens for certain packages(More about this in TODO  section)
- Write associated unit tests

#### TODOS and Open Questions
- [ ] Uncovered an issue with RM API in updating tokens at package level for certain packages when we provide the full payload containing coverage dates; which the team is aware of and will fix eventually at some point
- [ ] Revise and add more unit tests with full payload after the RM API team fixes the issue on their end.

## Screenshots
packagetokens

### DIFF
--- a/app/controllers/packages_controller.rb
+++ b/app/controllers/packages_controller.rb
@@ -103,6 +103,7 @@ class PackagesController < ApplicationController
         :isSelected,
         :allowKbToAddTitles,
         proxy: %i[id inherited],
+        packageToken: [:value],
         visibilityData: [:isHidden],
         customCoverage: %i[beginCoverage endCoverage]
       )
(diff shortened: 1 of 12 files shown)


carolegodfrey (approved):
Looks great @sduvvuri-ebsco - my only comment is if we should wait until rm api is corrected before making use of this functionality from the front end -- issue appears when setting token as part of a full payload

sduvvuri-ebsco:
I think it makes total sense to wait on the implementation in the UI until RM API fixes the issue on their end. However, we'll let Khalilah know and defer the decision to her since we are not sure when RM API will address the issue + it happens only on certain packages. We do not get any error, just that the update does not happen if we give the full payload, which is misleading. Will talk to Khalilah on Monday, so she can follow up on the timeline with Mohan.

PR #199, opened 2018-08-22
UIEH-531 Support nested search in PackageDetail Record

elrickvm:
## Purpose
Users want to be able to conduct additional search request from within a Package Detail Record that will filter down the list of titles / resources that is shown within a Package Detail Record.

A user conducts a package search from the main search. Clicks on an item from the results list. This opens a Package Detail record that contains another list of titles aka resources. 

To gather that list of title / resources a request is made behind the scenes to populate that list. The initial work to support searching for titles within a Package-Detail Record returned all the titles. This work will add the support to conduct a nested search to filter the title / resource list in a Package Detail record.

## Approach

Capture all the query params that are sent when a nested search or additional search is conducted from within a Package Detail record Titles list.

After capturing those query params we need to build the request and send it over to RMAPI that will then respond with the appropriate data. Then return the data to the Package Detail Record.


app/models/resource.rb
```diff
@@ -40,21 +40,42 @@ class Resource < RmApiResource
...
-          request.get_params[:q] ? 'relevance' : 'titlename'
+          request.get_params[:search].present? ? 'relevance' : 'titlename'
         end
 
-      request.get_params[:search] = request.get_params.delete(:q) || ''
-      request.get_params[:searchfield] = 'titlename'
       request.get_params[:count] ||= 25
       request.get_params[:offset] = request.get_params[:page] || 1
       request.get_params.delete(:page)
+      request.get_params.delete(:q)
```
sduvvuri-ebsco: Do we need this, specially now that we do not have user sending `q` param anymore?

spec/requests/packages_spec.rb
```diff
@@ -464,24 +463,32 @@
...
+            get '/eholdings/packages/19-6581/resources?'\
+                 'filter[name]=technology&sort=name',
                 headers: okapi_headers
           end
         end
 
         let!(:json_query_with_name_sort) { Map JSON.parse response.body }
 
-        it 'returns a list sorted by name' do
+        xit 'returns a list sorted by name' do
```
sduvvuri-ebsco: an additional `x` instead of `it`

sduvvuri-ebsco:
@elrickvm : have a couple of minor comments but otherwise, looks pretty good! Good to go as soon as those are addressed 👍

app/models/resource.rb
elrickvm: What i was noticing is that q is still on the request params. It is nil. But still present. So what happens is the request object is built to send to RMAPI but q is still on it as nil, unless you delete it.

Also, removing line 78 causes many test failures.

spec/requests/packages_spec.rb
elrickvm: xit skips the test and doesn't run the expectations. It leaves the test as `pending` in the test run. So future devs will know something is pending. I added a comment as to why it was skipped.

app/models/resource.rb
sduvvuri-ebsco: Well, I think this needs to be addressed in the UI, to not pass `q` probably? what does the complete request from UI look like? I am curious to see what's getting passed in the `q` if all we need is filter[xxx]?

spec/requests/packages_spec.rb
sduvvuri-ebsco: sounds good 👍

sduvvuri-ebsco (approved):
Looks good @elrickvm :shipit:

PR #200, opened 2018-08-28
Fix type of the `providerToken` field

MikeTaylor:
For some reason, the `type` of this field was an object containing an array: `{"type": ["object", "null"]}`. Unless there is some subtlety of JSON Schema that I don't know about, I think this must be a simple typo. At any rate, it prevented mod-graphql from properly ingesting the schema, so I have changed to to `"object"`, which seems from context to be correct.

(See https://issues.folio.org/browse/MODGQL-55 if interested in the history.)

### DIFF
--- a/ramls/providerListAttributes.json
+++ b/ramls/providerListAttributes.json
@@ -20,7 +20,7 @@
       "example": 49
     },
     "providerToken": {
-      "type": {"type": ["object", "null"]},
+      "type": "object",
       "description": "Provider token",
       "$ref": "providerToken.json"
     },


sduvvuri-ebsco (approved):
Change looks good @MikeTaylor! At some point, I vaguely remember, I was getting an error when the object was null instead of having a value in it, so I had to take this approach to get around that. But if you are not seeing any issues, I would say go ahead with the change specially that for now, we are not actually using these schemas for anything.

MikeTaylor:
Thanks!

PR #202, opened 2018-09-07
Fix proxy inheritance handling at package level to match provider and resource

sduvvuri-ebsco:
## Purpose
Carole noticed a bug while testing out proxy implementation of packages in ui-eholdings and on further analysis, we figured that resources was passing the exact same payload but isn't failing. So, identified the discrepancy in implementation between providers, packages and resources and fixed the packages portion of it.

## Screenshots
package_proxy


carolegodfrey (approved):
LGTM @sduvvuri-ebsco !

PR #204, opened 2018-09-10
Bump version up for further commits

sduvvuri-ebsco:
## Purpose
Since we released v1.0.0 last week, bump version up for further commits until next release.


funkymalc:
We actually need to get this merged in order to be able to build folio-snapshot because the latest module descriptor according to Okapi dependency resolution is v1.0.0 but there is no 1.0.0 artifact in the snapshot repositories.

cherewaty (approved):

PR #206, opened 2018-09-14
Correct Package List Serializer for Proxy

carolegodfrey:
## Purpose
While investigating an issue reported - error `Proxy 'id' must be present when 'inherited' is false` when selecting or de-selecting some specific packages -- I observed an issue with the package list returned from mod-kb-ebsco. Specifically package list results were returning a proxy object with null id and inherited indicator.

     "proxy": {
                "id": null
                "inherited": null
            },

https://issues.folio.org/browse/UIEH-555 

Upon further investigation determined that proxy id is not returned as an element from rm api list response. It is only available when retrieving a detailed record.

## Approach
Update serializable_package_list to remove proxy.
Update serializable_package to include proxy

## Screenshots
Before

After

### DIFF
--- a/app/serializable/serializable_package.rb
+++ b/app/serializable/serializable_package.rb
@@ -2,7 +2,8 @@
 
 class SerializablePackage < SerializablePackageList
   attributes :allow_kb_to_add_titles,
-             :package_token
+             :package_token,
+             :proxy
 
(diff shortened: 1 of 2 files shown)


sduvvuri-ebsco (approved):
Looks good Carole!

PR #205, opened 2018-09-10
MODKBEBSCO-6: Store rm api credentials in same location as mod-codex-ekb

carolegodfrey:
## Purpose
Per https://issues.folio.org/browse/MODKBEBSCO-6 mod-kb-ebsco should use the same location as mod-codex-ekb to store rm api credentials. 

mod-codex-ekb maintains 3 separate entities using mod-configuration to store url, customer id and api key.

```POST https://okapi.frontside.io/configurations/entries```

Url:

```{
    "module": "EKB",
    "configName": "api_access",
    "code": "kb.ebsco.url",
    "description": "EBSCO RM-API URL",
    "enabled": true,
    "value": "https://api.ebsco.io"
}
```
Customer Id: 
```{
    "module": "EKB",
    "configName": "api_access",
    "code": "kb.ebsco.customerId",
    "description": "EBSCO RM-API Customer ID",
    "enabled": true,
    "value": "apidvcorp"
}
```
API Key

```{
    "module": "EKB",
    "configName": "api_access",
    "code": "kb.ebsco.apiKey",
    "description": "EBSCO RM-API API Key",
    "enabled": true,
    "value": “XXXXXXXXXX”
}
```

mod-kb-ebsco maintains 1  entity to store customer id and api key and uses an environment variable to store url.

Using the same location provides the following benefits:

- By storing rm api URL in configuration (not environment variable) individual tenants can be configured to use either sandbox or production environments.

- Mod-codex-ekb can use ui-eholdings settings page to configure customer id, api key and url (once added to front end)

## Approach
- Remove use of environment variable `EBSCO_RMAPI_BASE_URL`
- Issue `GET /configurations/entries?query=module=EKB` instead of `GET /configurations/entries?query=module=KB_EBSCO` to get entries for customer id, url and api key
- Update configurations GET/PUT request to include added data attribute for RM API URL - `rmapiBaseUrl`
- Update configuration save to post request to new configuration location
- Correct not helpful `invalid base` error when entering incorrect credentials. Instead return `Invalid KB API Credentials`
- Re-recorded VCR(s) for unit tests. Majority of tests have GET interaction to obtain credentials. These have been re-recorded to reflect new endpoint
https://okapi.frontside.io/configurations/entries?query=module=EKB. Needed to adjust some tests due to differences in data from when the tests were originally run. Also skipped `xit` some titles test pending updates to make use of advanced searching in RMAPI

Sample `GET /eholdings/configuration`
```
{
    "data": {
        "id": "configuration",
[...]

PR #207, opened 2018-09-17
Add default handling for rm api url when not part of configurations PUT request

carolegodfrey:
## Purpose
After most recent update to mod-kb-ebsco https://github.com/folio-org/mod-kb-ebsco/pull/205, the configurations end  point PUT request now accepts rmapiBaseUrl as part of the request.

```{
    "data": {
        "id": "configuration",
        "type": "configurations",
        "attributes": {
            "customerId": "custid",
            "apiKey": "****************************************",
            "rmapiBaseUrl": "https://api.ebsco.io"
        }
    },
    "jsonapi": {
        "version": "1.0"
    }
}
```

UI eHoldings has not yet been updated to pass in the url parameter -- this will be covered in the following planned story.
https://issues.folio.org/browse/UIEH-548

In the interim, mod-kb-ebsco needs to support PUT Configurations requests from UI eHoldings that do not include a rmapiBaseUrl attribute yet

## Approach
Update configurations PUT request as follows:
(1) use rmapiBaseUrl attribute if it is passed as part of the PUT request
(2)use current configuration setting (if it is available)
(3)default to `https://sandbox.ebsco.io` if 1 and 2 are not available

### DIFF
--- a/app/controllers/configurations_controller.rb
+++ b/app/controllers/configurations_controller.rb
@@ -11,7 +11,7 @@ class ConfigurationsController < ApplicationController
 
     config.customer_id = data_attributes['customerId']
     config.api_key = data_attributes['apiKey']
-    config.rmapi_base_url = data_attributes['rmapiBaseUrl']
+    config.rmapi_base_url = get_base_url(data_attributes)
 
     if config.save
       render jsonapi: config,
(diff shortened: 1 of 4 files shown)


sduvvuri-ebsco (approved):
LGTM @carolegodfrey

PR #210, opened 2018-09-20
UIEH-559: Response should contain only identifier types and subtypes that UIEH needs

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-559, while running nightmare tests on UIEH, we were seeing flickering issue between list and detailed record. One of the reasons for that effect is that list records and detailed records contain differing lists of identifiers. Per our UI requirement, we need only those identifiers that have type = 0(ISSN) or 1(ISBN) and subtype = 1(Print) or 2(Online). This PR modifies the backend code to provide UI only with identifiers that are needed by UIEH.

## Approach
- In the place, where we are normalizing list to return sorted list of identifiers, filter the list that we get from RM API, sort the filtered list and return in response.
- Adjust unit tests accordingly because some of them look for other types or subtypes that mod-kb-ebsco no longer returns.

## Screenshots
issn-isbn

### DIFF
--- a/app/serializable/serializable_title_list.rb
+++ b/app/serializable/serializable_title_list.rb
@@ -43,7 +43,8 @@ class SerializableTitleList < SerializableJSONAPIResource
     }
 
     if @object.identifiersList
-      sorted_identifiers_list = @object.identifiersList.sort_by { |identifier| [identifier.subtype, identifier.type] }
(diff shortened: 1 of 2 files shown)


carolegodfrey (approved):
LGTM Sobha

PR #211, opened 2018-09-24
MODKBEBSCO-8 & MODKBEBSCO-10: Cleanup q param from titles endpoint as well as RAML

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/MODKBEBSCO-8, after refactoring param `q=` to be `filter[name]=` instead, there is some left over code in mod-kb-ebsco to be cleaned up. This PR addresses that cleanup. Also, per https://issues.folio.org/browse/MODKBEBSCO-10, we have to update `/eholdings/titles` endpoint documentation to reflect the same.

## Approach
- Code cleanup in `titles_controller.rb` and `title.rb` to not use `q` anymore
- Modify unit tests and ensure that they still pass as expected.
- Double check a few requests to ensure that they work as expected.
- Cleanup RAML documentation for `/eholdings/titles` endpoint and run raml-lint to ensure that its valid.

## Screenshots
code_cleanup

### DIFF
--- a/app/controllers/titles_controller.rb
+++ b/app/controllers/titles_controller.rb
@@ -14,7 +14,6 @@ class TitlesController < ApplicationController
 
     if title_query_params_validation.valid?
       @titles = titles.all(
-        q: params[:q],
         page: params[:page],
         filter: params[:filter],
         sort: params[:sort],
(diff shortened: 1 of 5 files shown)


carolegodfrey (approved):
Looks Good -- may also need to update API tests after this change is merged.

sduvvuri-ebsco:
Yep, correct @carolegodfrey! good point! Will update API tests as well.

PR #212, opened 2018-09-27
UIEH-551: Catch timeout exception in mod-kb-ebsco

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/UIEH-551, we are not capturing a timeout exception and displaying it correctly in UIEH. When we look at n/w traffic, we get a 500. 

## Approach
- My understanding is that the timeout exception needs to be caught in mod-kb-ebsco and formatted to use JSON API error response which can then be captured by UIEH and displayed accordingly.
- When we search for titles within package, mod-kb-ebsco was passing an extra non required parameter `q=` in the request to RM API which has been replaced by using `filter[name]=`, so remove this extra parameter and adjust vcr recordings accordingly because now the requests to RM API do not contain `q=` and vcr does not know how to handle those requests.
- Catch the timeout exception in `catch_flexirest_exceptions` and add associated middleware test

#### TODOS and Open Questions
- [x] Now, that we are catching the exception, we should check if the expected error response is being returned by mod-kb-ebsco. Haven't been able to check that yet since I was not able to reproduce 500 since. Update: Was able to test that this actually works.
- [ ] Once the above TODO has been verified, ensure that the error response is being handled and displayed correctly in UIEH - this might be a separate PR

## Learning
- Being able to come up with a test for this was pretty challenging - Since this is an exception in the middleware and not in the actual RM API code, could not use VCR recording to capture it. 
- Tried to see if Flexirest allows stub calls per here: https://github.com/flexirest/flexirest/blob/76c837491d1afa76b57b77b637969039bdd94bba/docs/faking-calls.md but that did not work either
- Tried using other forms of stubbing external calls but no luck: https://robots.thoughtbot.com/how-to-stub-external-services-in-tests
- https://github.com/bblimke/webmock
- Eventually, resorted to stubbing an instance method call instead of the HTTP request itself, using this https://relishapp.com/rspec/rspec-mocks/v/2-6/docs/method-stubs/stub-on-any-instance-of-a-class and ended up with deprecation warnings, which were fixed following this: https://relishapp.com/rspec/rspec-mocks/docs/working-with-legacy-code/any-instance

## Screenshots
request_timed_out


spec/requests/packages_spec.rb
```diff
@@ -553,6 +553,25 @@
     end
   end
 
+  describe 'getting resources related to package times out' do
+    before do
+      expect_any_instance_of(Faraday::Connection).to receive(:get).and_return(Map(body: { 'errors': [{ 'title': 'Request timed out', 'detail': 'Timed out getting https://api.ebsco.io/rm/rmaccounts/apidvcorp/vendors/19/packages/796/titles?count=25&offset=1&orderby=relevance&resourcetype=all&search=aa&searchfield=titlename&selection=all' }], 'jsonapi': { 'version': '1.0' } }.to_json, status: 408))
+    end
+
```
carolegodfrey, line 560: Nice to know how to mock this!

carolegodfrey (approved):
LGTM!

spec/requests/packages_spec.rb
sduvvuri-ebsco: Yes, it was pretty challenging. Just updated the `learning` part with all learning that I had to go through :)

PR #214, opened 2018-09-28
MODKBEBSCO-7: Update raml file and examples for missing token support

carolegodfrey:
## Purpose
Per https://issues.folio.org/browse/MODKBEBSCO-7 -- RAML needs to be updated to include documentation for tokens.

## Approach

- update packages PUT and GET requests to include packageToken and Proxy
- update provider GET response example to include an example Proxy
- ran ./scripts/lint-raml-cop.sh to verify raml
- execute raml2html eholdings.raml > eholdings.html to create new api docs locally

## Screenshots
screen shot 2018-09-28 at 4 39 57 pm

### DIFF
--- a/ramls/eholdings.raml
+++ b/ramls/eholdings.raml
@@ -347,6 +347,21 @@ types:
                         Note that this attribute can be updated both for CUSTOM AND MANAGED PACKAGES.
                       type: visibilityData
                       required: false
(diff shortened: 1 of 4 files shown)


sduvvuri-ebsco (approved):
Looks good!

